PR #1, opened 2023-06-26
Add workflows

PhoebeJackson1:
Adding workflows to check terraform


.github/workflows/terraform_security.yml
```diff
@@ -0,0 +1,18 @@
...
+    defaults:
+      run:
+        working-directory: terraform
+    env:
+      AWS_REGION: "eu-west-1"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: tfsec
+        uses: aquasecurity/tfsec-action@v1.0.0
```
BenRamchandani, line 18: Can you pass through the GITHUB_TOKEN as well and bump the version to 1.0.3? It occasionally rate limits otherwise.

See https://github.com/communitiesuk/delta-common-infrastructure/blob/main/.github/workflows/check.yml#L35, we should update it for Delta too at some point.

BenRamchandani (approved):
:+1:

PR #2, opened 2023-06-27
Hello world app

BenRamchandani:
Add a ktor app, this is mostly the default app generated by IntelliJ with a few plugins added and

* An SSL listener with a self-signed cert
* Docker build (ktor has a built in `buildImage` gradle task, but I figured we'd want to control it ourselves)
* GitHub action for building and running tests

I've used gradle as that's what it came with, does anyone have a preference there?

I haven't tried deploying it yet, I'll do that today.


auth-service/gradle.properties
```diff
@@ -0,0 +1,4 @@
+ktor_version=2.3.1
```
hugh-emerson, line 1: 2.3.2 released on Wednesday :P

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/SelfSignedSSLCertKeystore.kt
```diff
@@ -0,0 +1,33 @@
+package uk.gov.communities.delta.auth
+
+import io.ktor.network.tls.certificates.*
+import org.slf4j.LoggerFactory
+import java.io.File
+import java.security.KeyStore
+
+/*
+* We deploy behind an AWS ALB which doesn't validate SSL certificates anyway
+* so generate a self-signed certificate on startup
```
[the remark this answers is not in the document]
BenRamchandani: Pretty minimal, it offers some protection against a passive eavesdropper that can see VPC traffic. I was planning to let other apps continue connecting to the ALB like they are at the moment, which does have a valid cert.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import io.ktor.server.application.*
+import io.ktor.server.engine.*
+import io.ktor.server.netty.*
+import uk.gov.communities.delta.auth.plugins.configureMonitoring
+import uk.gov.communities.delta.auth.plugins.configureRouting
+import uk.gov.communities.delta.auth.plugins.configureSecurity
+import uk.gov.communities.delta.auth.plugins.configureSerialization
+
+fun main() {
```
hugh-emerson, line 11: What did you use to set up the project? Intellij's new ktor project?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/Routing.kt
```diff
@@ -0,0 +1,13 @@
+package uk.gov.communities.delta.auth.plugins
+
+import io.ktor.server.routing.*
+import io.ktor.server.response.*
+import io.ktor.server.application.*
+
+fun Application.configureRouting() {
```
hugh-emerson, line 7: How come there was no need to `install(Routing)`? From my very brief look at ktor docs it says to do that.

hugh-emerson (approved):
As part of setting up a skeleton app, we could also look into these now but fine by me if you're doing them as part of DT-481
* environment specific conf files
* Dependency injection (koin?)
* Project structure, like an example of a route defined in its own file because we aren't going to want to put everything in Routing.kt.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/Routing.kt
BenRamchandani: Looks like have a `routing {...}` block does it for you

https://ktor.io/docs/routing-in-ktor.html
> there is a convenient `routing` function that makes it simpler to install routing
> [...] `install(Routing)` is replaced with the `routing` function

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
BenRamchandani: Yep

hugh-emerson:
Also, yeah I'm happy with gradle. I prefer it to maven and some of the other projects already use gradle.

auth-service/gradle.properties
BenRamchandani: Updated, why not :D

BenRamchandani:
Will address the other points in DT-481

PR #3, opened 2023-06-30
DT-481 SAML token generation endpoint

BenRamchandani:
I've copied the code from the API (again) and converted it to Kotlin (or IntelliJ did anyway).

I've committed the private SAML signing key from the Delta repo that we use for development. I don't think there's any problem having this be public, but if you can think of one let me know.

For the code structure I've kept things pretty minimal:

* Config is environment variables and some hardcoded defaults in a static class
* Kept the "Routing.kt" file with the routes in, though most of the logic is in the controller
* A framework seemed over the top for three classes, so dependency injection is some static methods in Injection.kt

Running on test, can try it with

```sh
curl -X POST https://auth.delta.test.communities.gov.uk/auth-internal/generate-saml-token -u delta.app:Password123
```
from the bastion server


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/saml/SAMLTokenService.kt
```diff
@@ -0,0 +1,178 @@
...
+
+class SAMLTokenService(basicCredentials: BasicX509Credential) {
+    private val samlAssertionSigner: SAMLAssertionSigner
+    private val logger = LoggerFactory.getLogger(this.javaClass)
+
+    init {
+        samlAssertionSigner = SAMLAssertionSigner(basicCredentials)
+    }
+
+    fun generate(user: DeltaLdapPrincipal, validFrom: Instant, validTo: Instant): String {
```
hugh-emerson, line 33: Can we usefully unit test this?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/config/LDAPConfig.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.communities.delta.auth.config
+
+class LDAPConfig {
+    companion object {
+        val LDAP_URL = System.getenv("DELTA_LDAP_URL") ?: "ldap://localhost:2389"
```
hugh-emerson, line 5: README should mention the need for port forwarding localhost:2389 to say test environment's dluhctest.local

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    authenticate(DeltaADLdapAuthentication.NAME) {
+        route("/auth-internal") {
+            get("/auth-diag") {
+                val principal = call.principal<DeltaLdapPrincipal>(DeltaADLdapAuthentication.NAME)!!
+                call.respond(principal)
+            }
+            post("/generate-saml-token") {
+                generateSAMLTokenController.generateSAMLToken(call)
+            }
```
hugh-emerson, line 40: Would it be slightly nicer to declare `fun Route.generateSamlTokenRoutes()` in the Controller and then call `generateSamlTokenRoutes()` here?

Kinda moot when it's just a single route, but I it makes the controller a bit more self contained?

Like in https://ktor.io/docs/structuring-applications.html#group_by_file

hugh-emerson (approved):
Looks good. I couldn't make that POST through the bastion but I could run the app locally and it worked

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/saml/SAMLTokenService.kt
BenRamchandani: There's a test that checks the controller endpoint with fake LDAP security and makes sure a token is generated, but the generation code itself isn't meaningfully tested. We could add unit tests, but I don't think it's that valuable, whether it works with MarkLogic is the only thing that really matters and we can only do that in end to end testing.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
BenRamchandani: Definitely can, either as a top level function like that, or as a method on the controller `fun routes(route: Route) {...}`.

It's not a strong preference, but I think I prefer having all the routes in one file like it is now. It would get unwieldy with lots of routes, but for now it's quite neat.

BenRamchandani:
The curl command should be working again with some more changes to https://github.com/communitiesuk/delta-common-infrastructure/pull/317

BenRamchandani:
Merging so I can build on top of it, happy to come back to the tests + routing

PR #8, opened 2023-07-14
DT-522 LDAP login endpoint

BenRamchandani:
Implement a login endpoint. Required copying `ldapAuthenticate` from Ktor and adding some AD-specific error handling which I've copied from Delta's `MyActiveDirectoryLdapAuthenticationProvider` and tested some, but not all of the cases.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/PublicDeltaLoginController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/PublicDeltaLoginController.kt
@@ -1,5 +1,6 @@
 package uk.gov.communities.delta.auth.controllers
 
+import io.ktor.http.*
 import io.ktor.server.application.*
 import io.ktor.server.request.*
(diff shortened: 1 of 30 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/PublicDeltaLoginController.kt
```diff
@@ -1,18 +1,118 @@
...
         )
+    )
+
+    suspend fun loginPost(call: ApplicationCall) {
+        val formParameters = call.receiveParameters()
+        val formUsername = formParameters["username"]
+        val password = formParameters["password"]
+
+        if (formUsername.isNullOrEmpty()) return call.respondLoginPage(
+            errorMessage = "Username is required", errorLink = "#username"
```
BenRamchandani, line 41: Respond with the login page directly since we don't have sessions yet. Ideally also we'd display both username and password required if they're both empty, but I've kept it simple for now.

PR #10, opened 2023-07-25
DT-335 Signing Delta users in through Auth Service

BenRamchandani:
Includes changes for

* https://dluhcdigital.atlassian.net/browse/DT-518 Public routing
* https://dluhcdigital.atlassian.net/browse/DT-521 Set up database
* https://dluhcdigital.atlassian.net/browse/DT-522 Public login endpoint
* https://dluhcdigital.atlassian.net/browse/DT-524 Token endpoint

Still to do

* https://dluhcdigital.atlassian.net/browse/DT-523 Rate limiting
* https://dluhcdigital.atlassian.net/browse/DT-520 Public website setup
  * Catch all for CSP, cache control headers, CSRF etc.
* (No ticket) Configure sessions, for now we don't strictly need them in the auth service so I've skipped it
* (No ticket) Deleting expired rows from the database tables
  * Could simply always do this on writes to those tables, but ideally they'd be scheduled jobs which we don't have a standard way of doing currently (outside of MarkLogic...)
* Further changes on the Delta side, see https://github.com/communitiesuk/delta/pull/1136


auth-service/README.md
```diff
@@ -6,20 +6,28 @@
 * Forward localhost:2389 to port 389 of the test environment Active Directory,
   see the delta-common-infrastructure repository for details of the bastion host,
   and delta for port forwarding commands
+* Copy `.env.template` to `.env` and fill it in as instructed in that file
+* Postgres, by default on port 5438, use `docker compose up -d`
 
 ## Run
 
+* Load environment variables `export $(grep -v '^#' .env | xargs)`
```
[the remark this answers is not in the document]
BenRamchandani: Doesn't entirely make sense to me either, but seems to work and it's nice to have. IntelliJ ignores it though, still have to set it there.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
```diff
@@ -30,29 +28,31 @@ fun Route.healthcheckRoute() {
...
+    route("/delta") {
+        route("/login") {
+            get {
+                deltaLoginController.loginGet(call)
+            }
+            post {
+                deltaLoginController.loginPost(call)
             }
         }
     }
```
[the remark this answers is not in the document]
BenRamchandani: I've done a bit of both with functions in Routing.kt and get/post inside controllers

auth-service/README.md
```diff
@@ -6,20 +6,27 @@
...
 
 ## Run
 
 * Start with `./gradlew run`
 
 Alternatively open in IntelliJ, add this folder as a Gradle module, then create a new Ktor run configuration
-with `uk.gov.communities.delta.auth.ApplicationKt` as the main class.
+with `uk.gov.communities.delta.auth.ApplicationKt` as the main class and environment variables set from `.env`.
+
+Set the `io.ktor.development` property to `true` (`-Dio.ktor.development=true` JVM arg) to enable development mode (faster restarts, reloading of templates).
```
hugh-emerson, line 19: To enable same for command line, could you add it to build.gradle?

```
application {
    applicationDefaultJvmArgs = listOf("-Dio.ktor.development=true")
}
```

hugh-emerson (approved):

auth-service/build.gradle.kts
```diff
@@ -37,16 +37,50 @@ dependencies {
...
+
+// Migrations are run on startup outside of Development mode.
+// Extra task to run Flyway migrations without starting the
+task("migrate", JavaExec::class) {
+    mainClass.set("uk.gov.communities.delta.auth.services.DbPoolKt")
+    classpath = sourceSets["main"].runtimeClasspath
+}
+
+tasks {
+    "run"(JavaExec::class) {
```
hugh-emerson, line 74: Oh, is this equivalent to `task("run", JavaExec::class) {` like you used above?

BenRamchandani: Nope, that gives "Cannot add task 'run' as a task with that name already exists." 🤷

PR #13, opened 2023-08-04
Dt 523

PhoebeJackson1:
First half of the ticket - limiting and logging (metrics/alarms to follow)

Testing: 
- tested locally from a single user - gives helpful error message on log on page (also covered by tests)
- have not tested how it reads IP address via cloudfront and ALB (because not deployed yet) - will test this after merge - expected behaviour covered in tests
Risks - if implemented wrongly the login page could be attacked and we wouldn't notice up to the general WAF alarm limit
Documentation - [ticket](https://dluhcdigital.atlassian.net/browse/DT-523), this PR, will add changing the parameter value to the releases doc


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/StatusPages.kt
```diff
@@ -7,11 +7,20 @@ import io.ktor.server.plugins.statuspages.*
 import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.server.thymeleaf.*
+import uk.gov.communities.delta.auth.config.DeltaConfig
 
 private const val apiRoutePrefix = "/auth-internal/"
 
 fun Application.configureStatusPages(deltaWebsiteUrl: String) {
     install(StatusPages) {
+        // Currently only the login page is rate limited so always redirects to the login page for status "TooManyRequests"
```
BenRamchandani, line 16:
```suggestion
        // Currently only the login page is rate limited so always renders to the login page for status "TooManyRequests"
```
We're responding directly to the request, not redirecting

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/security/RateLimiting.kt
```diff
@@ -0,0 +1,35 @@
...
+import io.ktor.server.plugins.*
+import io.ktor.server.plugins.forwardedheaders.*
+import io.ktor.server.plugins.ratelimit.*
+import org.slf4j.LoggerFactory
+import uk.gov.communities.delta.auth.config.Env
+import kotlin.time.Duration.Companion.minutes
+
+fun Application.configureRateLimiting() {
+    val logger = LoggerFactory.getLogger("Application.RateLimiting")
+    val rateLimit = Env.getRequiredOrDevFallback("AUTH_RATE_LIMIT", "10").toInt()
```
BenRamchandani, line 13: Pass the rate limit in to the function rather than reading it here, that way we can change it in tests (so we don't have to make 10 requests, and they won't break if we change this default value)

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/security/RateLimiting.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    install(XForwardedHeaders) {
+        this.skipLastProxies(1)
+    }
+    install(RateLimit) {
+        // Currently only the login page is rate limited so a status 429 (Too many requests) sends the login page (with an error message)
+        register(RateLimitName("protectLogin")) {
```
BenRamchandani, line 20: Minor, but would be nice to have "protectLogin" as a public variable (`const val`) so other files can import it rather than needing the same string

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/security/RateLimiting.kt
BenRamchandani: Would be good to log the rate limit on startup too

BenRamchandani (approved):
Looks good 🙂

PR #18, opened 2023-08-15
DT-523

PhoebeJackson1:
Adding metrics and alarms for the auth service

Testing: Metrics tested by adding to the tests (all tests pass locally), tested alarms locally but will want further testing on test/staging once deployed (permissions for the role etc have not been tested)
Risks: Incorrectly set up alarms could mean that we aren't notified to login being spammed
Documentation: This PR, [the ticket](https://dluhcdigital.atlassian.net/browse/DT-523), will update the runbook with where to find the logs/metrics once deployed

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
@@ -1,6 +1,11 @@
 package uk.gov.communities.delta.auth
 
+import io.micrometer.cloudwatch2.CloudWatchConfig
+import io.micrometer.cloudwatch2.CloudWatchMeterRegistry
+import io.micrometer.core.instrument.Clock
+import io.micrometer.core.instrument.Counter
 import org.slf4j.Logger
+import software.amazon.awssdk.services.cloudwatch.CloudWatchAsyncClient
 import uk.gov.communities.delta.auth.config.*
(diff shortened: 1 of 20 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
```diff
@@ -73,6 +79,21 @@ class Injection (
     val ssoLoginStateService = SSOLoginSessionStateService()
     val ssoOAuthClientProviderLookupService = SSOOAuthClientProviderLookupService(azureADSSOConfig, ssoLoginStateService)
     val microsoftGraphService = MicrosoftGraphService()
+    val cloudWatchMeterRegistry = CloudWatchMeterRegistry(
```
BenRamchandani, line 82: We could check if `AUTH_METRICS_NAMESPACE` is set and use a SimpleMeterRegistry here instead if not

PhoebeJackson1: Done :) Please could you check it again since it seems to always work for me locally 🙄

BenRamchandani: That works 🙂

BenRamchandani (approved):
Looks good :)

PR #19, opened 2023-08-15
Remove password from returned pages

PhoebeJackson1:
Tested locally, password box is cleared on returning an LDAP error or incorrect password/login details error


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
```diff
@@ -120,7 +119,6 @@ class DeltaLoginController(
                         errorMessage = "Your account exists but is not set up to access Delta. Please contact the Service Desk.",
                         errorLink = deltaConfig.deltaWebsiteUrl + "/contact-us",
                         username = formUsername,
-                        password = password,
```
BenRamchandani, line 123: Should be able to remove the parameter from `respondLoginPage` too

BenRamchandani (approved):
:+1:

PR #20, opened 2023-08-15
Scheduled ECS tasks

BenRamchandani:
Updated, see comment below

---

Original text:

As a starting point for discussion.

Tasks that are scheduled to run in the background, and can also be started by a request.

The scheduled tasks are set to run every 12 hours from the point the app starts on a single threaded executor with no coordination, so they'll run once for each instance, which is fine in this case.

Other options could be:

* Moving the code for the tasks out into lambdas or ECS scheduled tasks independent from the auth service
* Using a lambda to hit the start task endpoint on a schedule
* Coordinating tasks in the database using e.g. <https://github.com/kagkarlsson/db-scheduler> or Quartz

If we're going to merge this:

* Add authentication of some kind to the new endpoint
* Unit tests
* Monitoring metrics


hugh-emerson (commented):
I mean, for a task like this, it's probably fine, but I don't think it's sufficient in general.
* It's running on every container
* You are setting the period rather than a fixed time
  * So we can't get it to run overnight to avoid impacting performance
  * If the container is replaced, the initial delay gets reset, right? So you if you released once a day, a daily job would never run. 

Of the options you listed, I'm leaning towards trying scheduled ECS tasks. 

How much effort do you think it would be to have a new image + task definition, which takes a "task" name when you run it and just runs that task and exits. I think it could be the same codebase but just with a change to the main function/entrypoint. Then some terraform code for an EventBridge Scheduler.

Running a task as a one off would mean going into ECS and running it?

I don't really want extra API endpoints if we don't need them.

BenRamchandani:
I've updated this to use ECS tasks and Event Bridge Scheduler.

I've added a new `main` function for it. The image and task definition are the same, but the scheduled one overrides the environment variable `RUN_TASK`, the entrypoint script picks that up and runs the new main function.

The AWS docs around it are a bit sparse, but it seems to work and it's pretty minimal extra infrastructure/code.

I've removed the scheduled thread pool executor and routes.

hugh-emerson (approved):
Looks good to me.

Are the scheduled tasks and their results clear enough in the AWS console?

The logs go to the main auth-service app log group? Would have been nice to have a separate log group but not a big deal - probably not worth creating a separate task definition just for that.

BenRamchandani:
> Are the scheduled tasks and their results clear enough in the AWS console?

There's no nice view of executions unfortunately 🙁 You can see the metrics and logs in CloudWatch, but that's it.

> The logs go to the main auth-service app log group? Would have been nice to have a separate log group but not a big deal - probably not worth creating a separate task definition just for that.

Yeah would definitely have been nicer, but there doesn't seem to be a way to do that with ContainerOverrides

PR #21, opened 2023-08-17
Dt 520 security headers

OliviaKRose:
Added caching and CSP headers

Testing: manually checked for headers on status page/login page

### DIFF
--- a/auth-service/build.gradle.kts
+++ b/auth-service/build.gradle.kts
@@ -51,6 +51,9 @@ dependencies {
     implementation("io.ktor:ktor-server-rate-limit:$ktorVersion")
     implementation("io.ktor:ktor-server-forwarded-header:$ktorVersion")
 
+    // Headers
+    implementation("io.ktor:ktor-server-caching-headers:$ktorVersion")
+
     // Metrics
     implementation("io.ktor:ktor-server-metrics-micrometer:$ktorVersion")
     implementation("io.micrometer:micrometer-registry-cloudwatch2:1.11.2")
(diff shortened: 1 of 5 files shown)


auth-service/build.gradle.kts
```diff
@@ -61,7 +64,8 @@ dependencies {
     // Database
     implementation("org.postgresql:postgresql:42.6.0")
     implementation("com.zaxxer:HikariCP:5.0.1") // Connection pool
-    implementation("org.flywaydb:flyway-core:9.20.1") // Migrations
+    implementation("org.flywaydb:flyway-core:9.20.1")
+    implementation("io.ktor:ktor-server-caching-headers-jvm:2.3.2") // Migrations
```
BenRamchandani, line 68: It sometimes does this when building, I'm not sure if it's IntelliJ trying to be helpful or what. If you add the "-jvm" bit to line 53 and delete this one it should be fine.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
```diff
@@ -1,22 +1,32 @@
...
 import uk.gov.communities.delta.auth.plugins.addServiceUserUsernameToMDC
 import uk.gov.communities.delta.auth.security.*
 
 fun Application.configureRouting(injection: Injection) {
     routing {
+        install(CachingHeaders) {
+            options { call, _ ->
+                if(call.response.headers["Cache-Control"] == null)
+                    CachingOptions(CacheControl.NoStore(CacheControl.Visibility.Private))
+                else null }
```
BenRamchandani, line 28: Don't worry here, but for next time worth formatting files, you can either do them as you go, it's ctrl + alt + L in IntelliJ, or before you commit you can right click on a folder and "Refactor code", and that should give you an option to refactor all changed files.

BenRamchandani (commented):
It looks like the CachingHeaders plugin can either be "installed" on a route or as a global plugin.
Currently you've got it in the `routing` block in Routing.kt as a route-level plugin, and in StatusPages.kt as a global plugin.

I think we just want the global one. Can you take the `if(call.response.headers["Cache-Control"] == null)` check from Routing.kt and put it in the one in StatusPages.kt? Then let's move it to its own file.

BenRamchandani (approved):
Looks good, thank you 🙂 

I've just merged a load of changes (sorry!), let me know if you need a hand with any of the merge conflicts

PR #23, opened 2023-08-21
Use IO dispatcher for blocking calls

BenRamchandani:
Plus a couple of refactors (review commit by commit).

Something I've been lazy about and am coming back to now. From some reading and experimenting:

* Ktor uses Dispatchers.Default for handling requests, which will be 2 threads on prod (it's `max(cpu_cores, 2)`)
* JDBC and Java's LDAP implementation are blocking, so send those off to Dispatchers.IO (which is a bigger thread pool)
* HttpClient (for talking to Microsoft in the OAuth flow) is non-blocking, so don't need to worry about that

I haven't done any load testing as I expect we've got plenty of headroom even without this change.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
@@ -9,6 +9,8 @@ import io.ktor.server.routing.*
 import io.ktor.server.sessions.*
 import io.ktor.server.thymeleaf.*
 import io.micrometer.core.instrument.Counter
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 import org.slf4j.LoggerFactory
 import uk.gov.communities.delta.auth.LoginSessionCookie
 import uk.gov.communities.delta.auth.config.AuthServiceConfig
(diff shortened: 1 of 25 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
```diff
@@ -162,9 +164,11 @@ class DeltaLoginController(
                     )
                 }
 
-                val authCode = authenticationCodeService.generateAndStore(
-                    userCn = loginResult.user.cn, client = client, traceId = call.callId!!
-                )
+                val authCode = withContext(Dispatchers.IO) {
```
hugh-emerson, line 167: Is it not possible to put the `withContext(Dispatchers.IO)` inside the `generateAndStore` function? Does it need done at the interface like this?

hugh-emerson (approved):

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
BenRamchandani: Yep could, the reason I didn't do that is because the other function (lookupAndInvalidate) was already called from an IO thread. I don't think doing a nested `withContext(Dispatchers.IO)` is really an, I'll maybe try switching to that.

PR #24, opened 2023-08-22
Remove redundant count ternary - avoids  Invalid count argument error

hugh-emerson:
No change to resources, just changes how state is stored.

PR is just for visibility - this has been applied to prod.


BenRamchandani (approved):
Thanks :+1:

PR #27, opened 2023-08-24
DT-570

PhoebeJackson1:
Move register page, add errors and handling, set up controller, plan rest of ticket

Progress on ticket so far and lots of TODOs describing expected plan whilst @hugh-emerson and @BenRamchandani are on holiday - please suggest any potential issues if you can see them/ask for further clarification

-----------------

A few less todos - emails now send, users are created and added to groups/organisation. Still to go - SSO User account creation flow, set password flow, error handling/logging improvements, retired org stuff,  few bits and bobs

------------------

Tests still to come, a few questions in TODOs
Not sure if this is safe from timing attacks on the comparisons of tokens @BenRamchandani ?
Still need to set the environment variable values for non local environments
Due to not moving the admin user creation flow as part of this ticket the aim of the ticket to "remove the link between AD and delta" has not been fully achieved. Also since users created that way will be given password set tokens from delta the vulnerability in DT-393 is also not resolved since that code is still utilised.

-----------------

Now includes all tests :) 
Still need to set the environment variable values for non local environments
Corresponding delta PR with updated playwright tests and other delta side changes to come

----------------

New metrics alarms not tests but structured the same as existing ones so low risk
Environment variables set
Delta PR raised (https://github.com/communitiesuk/delta/pull/1236)


auth-service/build.gradle.kts
```diff
@@ -58,6 +58,9 @@ dependencies {
     implementation("io.ktor:ktor-server-metrics-micrometer:$ktorVersion")
     implementation("io.micrometer:micrometer-registry-cloudwatch2:1.11.2")
 
+    //Emails
+    implementation("com.sun.mail:javax.mail:1.6.2")
```
BenRamchandani, line 62: https://mvnrepository.com/artifact/com.sun.mail/jakarta.mail is the modern version

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/Templating.kt
```diff
@@ -2,24 +2,32 @@ package uk.gov.communities.delta.auth.plugins
...
+    }
+}
+
+fun makeTemplateResolver(devMode: Boolean): AbstractConfigurableTemplateResolver {
+    val logger = LoggerFactory.getLogger("Application.Templating")
+
+    return (if (devMode) {
+        logger.info("Development mode, using templates from src/ folder")
+        FileTemplateResolver().apply {
+//            cacheManager = null // TODO - add this back in?
```
BenRamchandani, line 22: Did it break something? (we don't want to cache them in dev mode so changes get picked up quickly)

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/EmailService.kt
```diff
@@ -0,0 +1,47 @@
...
+        subject: String,
+        mappedValues: Map<String, String>,
+    ) {
+        val templateEngine = TemplateEngine()
+        templateEngine.setTemplateResolver(makeTemplateResolver(false))
+        val context = Context(Locale.getDefault(), mappedValues)
+        val content = templateEngine.process(template, context)
+        try {
+            println("Emails not set up yet but will send an email to $toEmail with subject $subject and content $content from $fromEmail ($fromName) with replies going to $replyToEmail ($replyToName)")
+//            TODO - set this up to send emails
```
BenRamchandani, line 32: Will involve passing in configuration and credentials similar to Delta. I think it's fine for it to use the same credentials as Delta itself.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/RegistrationService.kt
```diff
@@ -0,0 +1,155 @@
...
+        constructor(
+            status: Status,
+            email: String,
+            token: String
+        ) {
+            this.status = status
+            this.email = email
+            this.token = token // TODO - are these still needed with tokens stored in db (see below)
+        }
+    }
```
BenRamchandani, line 39: You can do something like

```kotlin
sealed class RegistrationResult
class ResultCreated(val email: String, val token: String): RegistrationResult
class ResultAlreadyExists(val email: String): RegistrationResult
class ResultFailure(val exception: Exception?): RegistrationResult
```

then

```kotlin
when (result) {
  is ResultCreated -> { doSomethingWith(result.token) }
  ...
}
```

`sealed` means the compiler knows there aren't any other instances of the class so it can check the `when` is exhaustive

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/RegistrationService.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    //        val context = Context(Locale.getDefault(), model) TODO - what did this do? Is it still needed? - from email sending
+
+
+    fun register(registration: Registration): RegistrationResult { // TODO - does this still need to return anything?
+        val user = userService.findUserById(registration.emailAddress, true).orElse(null)
+        if (user != null) {
+            // TODO - logic check - do we want this to happen?
+            //  - if a user exists already then registering them again will add them to organisations associated with their domain
+            //      - can't see a scenario where this isn't already the case but if it is then do we definitely want to fix it?
```
BenRamchandani, line 58: I don't think so, should just say the account exists and send them to the login page. It kinda made sense when there were non-delta users in datamart, but not any more.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/RegistrationService.kt
```diff
@@ -0,0 +1,155 @@
...
+//            deltaUserEmailService.sendEmail(user::getEmail)
+            return RegistrationResult(
+                RegistrationResult.Status.ALREADY_EXISTS,
+                registration.emailAddress
+            )
+        }
+//        TODO - implement creation of users as before (ish) for non-SSO
+//             - for SSO users (add domain check) - implement different create user route (no need for password)
+//                    - does this need to come via register page in the same way?
+//                    - is their account just created automatically on first log in or do they still have to register?
```
BenRamchandani, line 70: I think ideally it would send them to a pre-filled version of the register page, then if they e.g. already have an account under another email they can realise and try and sort it. If it's easier to just do it in the same request I think that's fine for now though.

And I don't think we need to make them validate their email.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/RegistrationService.kt
```diff
@@ -0,0 +1,155 @@
...
+//        TODO - add new user to necessary organisations (same for SSO creation and not-SSO creation)
+//        registrationOrganisationService.attachUserToDomainOrganisations(registration)
+
+//        TODO - get value for setPasswordUrl (if not SSO person)
+//              - create table in database for reset password tokens
+//                  - containing: user CN, token, timestamp
+//                  - watch for timing attacks on comparison of tokens
+//              - add row to table containing token
+//              - send email containing url as before (ensure url encoding happens where necessary)
+//              - compare token to tokens in the database on set password
```
BenRamchandani, line 82: and make sure each token can only be used once

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/RegistrationService.kt
```diff
@@ -0,0 +1,155 @@
...
+            mapOf(
+                "deltaUrl" to deltaConfig.deltaWebsiteUrl,
+                "userFirstName" to registration.firstName,
+//                    "setPasswordUrl" to setPasswordUrl
+            )
+        )
+        return RegistrationResult(
+            RegistrationResult.Status.CREATED,
+            registration.emailAddress,
+//            result.getToken() - TODO - what was this used for? Is it the password reset token
```
BenRamchandani, line 109: I was assuming so!

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/UserService.kt
```diff
@@ -0,0 +1,153 @@
...
+//              - createUser
+//              - any they (^) use
+
+    //    private val logger = LoggerFactory.getLogger(javaClass)
+//
+//    fun findUserById(id: String): Optional<User> {
+//        return findUserById(id, false)
+//    }
+//
+    fun findUserById(id: String?, includeNonDeltaUsers: Boolean): Optional<User> {
```
BenRamchandani, line 17: Delta wasn't, but we should be explicit, is this CN or email we're searching by?

auth-service/src/main/resources/templates/thymeleaf/delta-login.html
```diff
@@ -68,7 +68,7 @@ <h2 class="govuk-error-summary__title">
 
             <div class="govuk-grid-column-one-third">
                 <h2 class="govuk-heading-m">Not a user?</h2>
-                <a th:href="${deltaUrl} + '/register'" role="button" draggable="false"
+                <a th:href="'/delta/register'" role="button" draggable="false"
```
BenRamchandani, line 71: Should put this behind feature flag (i.e. environment variable)

auth-service/src/main/resources/templates/thymeleaf/new-user-email.html
```diff
@@ -0,0 +1,68 @@
...
+<p>Please click on the link below to set your password and activate your account.</p>
+
+<p>
+    <a th:href="@{${setPasswordUrl}}">Activate your DELTA account</a>. <b>This link is only valid for a short time so
+    please activate your
+    account now</b>.
+</p>
+<p>
+    Please note the DELTA User Account <a th:href="@{${deltaUrl}+'privacy-policy'}">Privacy Policy</a> and the
+    application <a th:href="@{${deltaUrl}+'usage-policy'}">Usage Policy</a>.
```
BenRamchandani, line 43: At least from DeltaConfig, the delta url doesn't include a trailing slash

auth-service/src/main/resources/templates/thymeleaf/registration-success.html
```diff
@@ -0,0 +1,32 @@
...
+        <h1 class="govuk-heading-xl">Registration pending</h1>
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <div class="govuk-panel">
+                    <h1>
+                        Your registration is pending your email activation
+                    </h1>
+                </div>
+
+                <p class="govuk-body">An email has been sent to your email address. Follow the instructions in the email to complete your registration. If you have not received an email after 15 minutes, contact the DLUHC Service Desk: <a class="govuk-link" href="mailto:dluhc.digital-services@levellingup.gov.uk">dluhc.digital-services@levellingup.gov.uk</a> or call 0203 829 0743.</p>
```
BenRamchandani, line 25: We should include the user's email here if it's easy to do so they can spot any typos

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/Templating.kt
PhoebeJackson1: It just didn't recognise itself as an option 🤷 😂

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/RegistrationService.kt
PhoebeJackson1: Yay will delete

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/RegistrationService.kt
PhoebeJackson1: Discussion in person concluded:
If people log in via SSO and don't have a delta account yet then we make them a delta account and redirect them to delta logged into that account
If people attempt to register for delta with an email that is one of the SSO required org domains then we redirect them to SSO and once they have logged in treat it as above

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/RegistrationService.kt
```diff
@@ -0,0 +1,147 @@
...
+) {
+    sealed class RegistrationResult
+    class UserCreated(val registration: Registration, val token: String) : RegistrationResult()
+    class UserAlreadyExists(val registration: Registration) : RegistrationResult()
+    class RegistrationFailure(val exception: Exception?) : RegistrationResult()
+
+    // Previously users could be datamart users but not delta users, at migration we only transferred over users who
+    // were delta users so there should never be users who exist but aren't delta users - therefore can remove all logic
+    // for attaching users and checking if they are indeed delta users (if they somehow existed they'd get a "contact
+    // the service desk" error at login which is what we want
```
BenRamchandani, line 28:
```suggestion
    // for attaching users and checking if they are indeed delta users (if they somehow existed they'd get a "contact
    // the service desk" error at login which is what we want)
```

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaUserRegistrationController.kt
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private val firstNameEmpty = "First name must not be empty"
+    private val lastNameEmpty = "Last name must not be empty"
+    private val emailAddressEmpty = "Email address must not be empty"
+    private val confirmEmailAddressEmpty = "Confirm email address must not be empty"
+    private val notEqualEmails = "Email addresses do not match"
+    private val notAnEmailAddress = "Email address must be a valid email address"
+    private val notAKnownDomain =
+        "Email address domain not recognised, please use an email address associated with an organisation using DELTA"
```
BenRamchandani, line 63: Should say something about contacting the service desk if they think it should work

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/ADGroup.kt
```diff
@@ -0,0 +1,21 @@
+package uk.gov.communities.delta.auth.utils
+
+import uk.gov.communities.delta.auth.config.LDAPConfig
+import javax.naming.directory.Attribute
+import javax.naming.directory.BasicAttribute
+
+class ADGroup(val cn: String, private val ldapConfig: LDAPConfig){
```
BenRamchandani, line 7: Same for ADUser, I think I'd make this an inner class of one of the services, it feels like an internal helper class and not something we should be returning back to controllers or other parts of the code. If it's here other code might start using it that shouldn't,

auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
```diff
@@ -44,6 +44,7 @@
                     </span>
                 </span>
         <span class="govuk-header__product-name">DELTA</span>
+<!--        TODO - extra space is annoying-->
```
BenRamchandani, line 47: 😄

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
```diff
@@ -66,11 +72,37 @@ fun Route.externalRoutes(
...
         deltaLoginRoutes(serviceConfig, deltaLoginController, deltaSSOLoginController)
     }
 }
 
+fun Route.deltaSetPasswordRoutes(deltaSetPasswordController: DeltaSetPasswordController) {
+    route("/success") {
+        deltaSetPasswordController.setPasswordSuccessRoute(this)
+    }
+    // TODO - should this be rate limited?
+    deltaSetPasswordController.setPasswordFormRoutes(this)
```
BenRamchandani, line 93: Wouldn't hurt, though there shouldn't be any practical attacks on it. The register endpoint definitely should be though.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaSetPasswordController.kt
```diff
@@ -0,0 +1,199 @@
...
+        val userCN = call.request.queryParameters["userCN"].orEmpty()
+        val token = call.request.queryParameters["token"].orEmpty()
+        when (setPasswordTokenService.useToken(token, userCN)) {
+            is SetPasswordTokenService.NoSuchToken -> {
+                throw SetPasswordException("set_password_no_token", "Set password token did not exist")
+            }
+
+            is SetPasswordTokenService.ExpiredToken -> {
+                sendNewSetPasswordLink(userCN)
+                call.respondExpiredTokenPage()
```
BenRamchandani, line 57: Sorry to make more work, but can you make sending a new activation email a button on the expired token page and have a separate endpoint for it? I think that will be clearer, and don't like side effects as part of GET requests

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaSetPasswordController.kt
```diff
@@ -0,0 +1,199 @@
...
+
+
+        val message: String? =
+            if (Strings.isNullOrEmpty(newPassword)) "New password is required."
+            else if (Strings.isNullOrEmpty(confirmPassword)) "Confirm password is required."
+            else if (newPassword != confirmPassword) "Passwords did not match."
+            else if (newPassword.length < 12) "Password must be at least 12 characters long."
+            else if (passwordChecker.isCommonPassword(newPassword, userCN))
+                "Passwords must not be a commonly used password format or contain your username"
+            else null
```
BenRamchandani, line 91: I think if there are form validation errors let's return an error before checking the token.

I can see why you've done it like this, but I think that would be easier to understand.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaSetPasswordController.kt
```diff
@@ -0,0 +1,199 @@
...
+
+        val errorPresent = message != null
+        val tokenResult = setPasswordTokenService.useToken(token, userCN, !errorPresent)
```
BenRamchandani, line 94: I think it would be clearer to have separate functions for validating a token and consuming one, even if they call the same private function in the service

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaSetPasswordController.kt
```diff
@@ -0,0 +1,199 @@
...
+                val userDN = String.format(ldapConfig.deltaUserDnFormat, tokenResult.userCN)
+                try {
+                    userService.setPassword(userDN, newPassword)
+                } catch (e: Exception) {
+                    logger.error("Error setting password for user with DN {}", userDN, e)
+                    return this.respondSetPasswordPage(
+                        tokenResult.userCN,
+                        tokenResult.token,
+                        "An error occurred please try again or contact the service desk"
+                    )
```
BenRamchandani, line 158: I think rethrow and use the generic error page here, that way we get a request id

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaSetPasswordController.kt
```diff
@@ -0,0 +1,199 @@
...
+
+    private suspend fun ApplicationCall.respondSetPasswordPage(
+        userCN: String,
+        token: String,
+        message: String? = null,
+    ) {
+        val mapOfValues = mutableMapOf(
+            "deltaUrl" to deltaConfig.deltaWebsiteUrl,
+            "userCN" to userCN,
+            "token" to token,
```
BenRamchandani, line 177: do we use this in the form?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/PasswordChecker.kt
```diff
@@ -0,0 +1,37 @@
...
+    private fun hasInsufficientUniqueCharacters(password: String): Boolean {
+        val uniqueCharacters: MutableSet<Char> = HashSet()
+        for (c in password.toCharArray()) uniqueCharacters.add(c)
+        return uniqueCharacters.size < 5
+    }
+
+    private fun containsPartOfUserEmail(lowerCasePassword: String, userEmail: String): Boolean {
+        val wordsInUserEmail =
+            userEmail.lowercase(Locale.getDefault()).split("[\\W0-9_]".toRegex()).dropLastWhile { it.isEmpty() }
+                .toTypedArray()
```
BenRamchandani, line 24:
```suggestion
            userEmail.lowercase(Locale.getDefault()).split("[\\W0-9_]".toRegex())
```
I think it should work without those

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/PasswordChecker.kt
```diff
@@ -0,0 +1,37 @@
...
+import java.util.*
+
+class PasswordChecker {
+
+    private fun containsCommonString(lowerCasePassword: String): Boolean {
+        val commonStrings = arrayOf("qwerty", "abcd", "123456", "password")
+        for (commonString in commonStrings) {
+            if (lowerCasePassword.contains(commonString)) return true
+        }
+        return false
```
BenRamchandani, line 12:
```suggestion
        listOf("qwerty", "abcd", "123456", "password").any { lowerCasePassword.contains(it) }
```

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/PasswordChecker.kt
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    private fun hasInsufficientUniqueCharacters(password: String): Boolean {
+        val uniqueCharacters: MutableSet<Char> = HashSet()
+        for (c in password.toCharArray()) uniqueCharacters.add(c)
+        return uniqueCharacters.size < 5
```
BenRamchandani, line 18:
```suggestion
        password.toCharArray().distinct().size < 5
```

auth-service/src/main/resources/db/migration/V2__password_tokens.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE set_password_tokens
+(
+    user_cn    text UNIQUE PRIMARY KEY,
+    token      bytea     NOT NULL,
+    created_at timestamp NOT NULL
+);
+
+CREATE INDEX set_password_token ON set_password_tokens (token);
```
BenRamchandani, line 8: Do we do any lookups on this?

auth-service/src/main/resources/db/migration/V2__password_tokens.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE set_password_tokens
+(
+    user_cn    text UNIQUE PRIMARY KEY,
```
BenRamchandani, line 3:
```suggestion
    user_cn    text      PRIMARY KEY,
```
primary key implies unique

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaUserRegistrationController.kt
```diff
@@ -0,0 +1,241 @@
...
+                    deltaRouteWithEmail(
+                        deltaConfig.deltaWebsiteUrl,
+                        ssoClientMatchingEmailDomain.internalId,
+                        emailAddress
+                    )
+                )
+            }
+
+            val registration = Registration(firstName, lastName, emailAddress)
+            lateinit var registrationResult: RegistrationService.RegistrationResult
```
BenRamchandani, line 132:
```suggestion
            val registrationResult: RegistrationService.RegistrationResult
```
Kotlin should be able to figure out it's always assigned exactly once before it's used

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaUserRegistrationController.kt
```diff
@@ -0,0 +1,241 @@
...
+                    lastName,
+                    emailAddress,
+                    e
+                )
+                throw e // TODO - handle errors in a tidier way so these two can be combined - make specific exceptions
+            }
+            try {
+                registrationService.sendRegistrationEmail(registrationResult)
+            } catch (e: Exception) {
+                // If it fails on this step the user never gets a link and doesn't know if account is made - TODO - is there a solution to this?
```
BenRamchandani, line 148: Not really 🤷 we could queue emails and retry them, but not worth the trouble with the small amount we send

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
```diff
@@ -0,0 +1,136 @@
...
+
+        val enabled = adUser.userAccountControl == ADUser.accountFlags(true)
+        if (enabled && adUser.password == null) {
+            logger.error("Trying to create an enabled user without a password")
+            throw UserVisibleServerError(
+                "passwordless_user",
+                "Trying to create user with no password",
+                "Something went wrong",
+                "Registration Error"
+            )
```
BenRamchandani, line 46: This should never happen right, it would be a bug? In which case can save a few lines and throw a regular `Exception` without logging anything first

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
```diff
@@ -0,0 +1,136 @@
...
+            fun accountFlags(enabled: Boolean): String {
+                return if (enabled) {
+                    NORMAL_ACCOUNT_FLAG.toString()
+                } else {
+                    return (NORMAL_ACCOUNT_FLAG + ACCOUNTDISABLE_FLAG).toString()
+                }
+            }
+
+            fun getPasswordAttribute(password: String): Attribute {
+                lateinit var bytes: ByteArray
```
BenRamchandani, line 124: Same thing, shouldn't need `lateinit var` here

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
```diff
@@ -99,7 +134,13 @@ fun Route.deltaLoginRoutes(
     }
 }
 
+fun deltaRouteWithEmail(deltaUrl: String, ssoClientInternalId: String, email: String) =
+    deltaUrl + "/oauth2/authorization/delta-auth?sso-client=${ssoClientInternalId}&expected-email=${email.encodeURLParameter()}"
+
 fun oauthClientLoginRoute(ssoClientInternalId: String) = "/delta/oauth/${ssoClientInternalId}/login"
+fun oauthClientLoginRouteWithEmail(ssoClientInternalId: String, email: String) =
+    "/delta/oauth/${ssoClientInternalId}/login?expected-email=${email.encodeURLParameter()}"
```
BenRamchandani, line 142: Could make this one function with a default parameter i.e. `oauthClientLoginRouteWithEmail(ssoClientInternalId: String, email: String? = null)` (not saying that would be better, just making sure you know it's an option!)

auth-service/src/main/resources/templates/thymeleaf/set-password-success.html
```diff
@@ -0,0 +1,34 @@
...
+        <h1 class="govuk-heading-xl">Registration completed</h1>
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-two-thirds">
+                <div class="govuk-panel">
+                    <h1>
+                        Your password has been set and your registration is completed
+                    </h1>
+                    <a th:href="${deltaUrl}" role="button" draggable="false"
+                       class="govuk-button govuk-button--secondary">
+                        Log in to DELTA
```
BenRamchandani, line 25:
```suggestion
                    <a th:href="${deltaUrl}/login" role="button" draggable="false"
                       class="govuk-button govuk-button--secondary">
                        Sign in to DELTA
```

For consistency with the error page which has a similar button

auth-service/src/main/resources/templates/thymeleaf/set-password-form.html
```diff
@@ -0,0 +1,79 @@
...
+                <div class="govuk-form-group"
+                     th:classappend="${message != null and !#strings.contains(message, 'Confirm password')} ? 'govuk-form-group--error' : ''">
+                    <label class="govuk-label" for="newPassword">New password</label>
+                    <div id="newPasswordError">
+                        <span class="govuk-visually-hidden">Error:</span>
+                        <p class="govuk-error-message"
+                           th:if="${message != null and !#strings.contains(message, 'Confirm password')}"
+                           th:text="${message}"></p>
+                    </div>
+                    <input id="newPassword" autocomplete="off" class="govuk-input govuk-!-width-two-thirds"
```
BenRamchandani, line 52:
```suggestion
                    <input id="newPassword" autocomplete="new-password" class="govuk-input govuk-!-width-two-thirds"
```
and I think the confirm box should be the same

BenRamchandani (commented):
Worth running IntelliJ's reformat code action over all the Kotlin files (commit everything first in case it makes a mess)

General point for logging for future reference), this isn't something we've stuck to religiously, but it's easier to search through logs when the message is a fixed string and everything else is a parameter, so instead of

```kotlin
logger.info("Created a user called {} in state {}", username, state);
```
do
```kotlin
logger.atInfo().addKeyValue("username", username).addKeyValue("userState", state).log("User created");
```

It's not worth the trouble everywhere (and don't go changing it now!), but the second version means you can do `fields username | filter message = "User created"` in Logs Insights easily without faffing around with `like`, so it's nice for anything you might want to get a list of.

The auth service is definitely set up to log those key values, I'm not sure Delta/CPM is, so check first if you're using it there.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaUserRegistrationController.kt
PhoebeJackson1: It doesn't like this because I've changed the error handling slightly and use it there so it thinks it could be not initialized although it can't be

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/EmailAddressChecker.kt
```diff
@@ -0,0 +1,23 @@
...
+import uk.gov.communities.delta.auth.services.Organisation
+import java.util.*
+
+
+class EmailAddressChecker {
+
+    fun hasValidEmailFormat(email: String): Boolean {
+        if (email.count { it == '@' } != 1) return false
+        val emailRegex = """^[\w-+.]+@([\w-]+\.)+[\w-]{2,4}$""".toRegex()
+        return emailRegex.matches(email)
```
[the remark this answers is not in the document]
PhoebeJackson1: Should I be using only that/storing the email regex in the same place to make it easier if any issue like the ' thing comes up again (I've also added ' to the email regex now), also it currently allows + s but the username one doesn't - do we allow them or should I remove it?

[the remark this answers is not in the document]
BenRamchandani: Wouldn't hurt to have them in the same place, and make it clear one is for signups and the other for logins. Fine to add `+` to the username one :+1:

auth-service/src/main/resources/templates/thymeleaf/register-user-form.html
```diff
@@ -0,0 +1,132 @@
...
+<!--/*@thymesVar id="emailAddress" type="java.lang.String" */-->
+<!--/*@thymesVar id="confirmEmailErrorMessages" type="java.lang.ArrayList" */-->
+<!--/*@thymesVar id="confirmEmailAddress" type="java.lang.String" */-->
+
+<div class="govuk-width-container" id="registerPage">
+    <ol class="govuk-breadcrumbs__list">
+        <li class="govuk-breadcrumbs__list-item"><a class="govuk-breadcrumbs__link" data-th-href="${deltaUrl}">Home</a>
+        </li>
+        <li class="govuk-breadcrumbs__list-item">Register</li>
+    </ol>
```
PhoebeJackson1, line 29: @BenRamchandani is this ok as is or should I look into the CSS? How it looks on new register page vs how it look on a page on delta
image
image

PhoebeJackson1: Also the extra space between Gov.uk and Delta?

BenRamchandani: I think that's fine

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/EmailAddressChecker.kt
```diff
@@ -7,16 +7,13 @@ import java.util.*
...
 class EmailAddressChecker {
 
-    fun hasValidEmailFormat(email: String): Boolean {
+    fun hasValidFormat(email: String): Boolean {
         if (email.count { it == '@' } != 1) return false
-
-        val emailRegex = """^[\w-+.']+@([\w-']+\.)+[\w-]{2,4}$""".toRegex()
         val cn = email.replace("@", "!")
-        return emailRegex.matches(email) && LDAPConfig.VALID_USERNAME_REGEX.matches(cn)
+        return LDAPConfig.VALID_EMAIL_REGEX.matches(email) && LDAPConfig.VALID_USERNAME_REGEX.matches(cn)
```
hugh-emerson, line 13: Really minor points, feel free to ignore me:

Is it even possible to match the email regex and not the username regex?

Also, not escaping the hyphen feels slightly risky - it possibly works currently because it's currently next to `\w`, but it can be a special character inside `[]`, denoting a range of characters, like `a-z`

hugh-emerson (commented):
Only unresolved comment is Ben's from two weeks ago - moving the resend email to a separate action. Is there a reason we don't want to do that?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/EmailAddressChecker.kt
PhoebeJackson1: I don't think it is possible to match email but not username but it is possible to match username but not email, but possibly both is useful incase a requirement changes and is updated in one place but not the other (although they are now in the same file so that is harder to do)

hugh-emerson: There's a user called "admin" on staging which is shared by some DLUHC folks for testing. And there's our various test users in the test environment. But the normal sign up process isn't possible for those anyway so I don't think we need to worry about allowing it.

BenRamchandani: `ADLdapLoginService` is also used for doing service user logins (like delta.app), so don't want to make it more restrictive. I think it's probably fine to check both

PhoebeJackson1: (resolving conversation, will rename it next time that stuff is changed)

BenRamchandani (approved):

Perrett411 (commented):
class AI:
    def __init__(self, ecosystem):
        self.knowledge = None
        self.skills = None
        self.ecosystem = ecosystem

# Define a function to load information into the AI
def load_information(self, data):
    self.knowledge = data

# Define a function to retrieve information
def retrieve_information(self, request):
    return self

PR #30, opened 2023-09-26
DT-553 Map user roles from AD

BenRamchandani:
Map the user's groups into a list of Delta roles and access groups.

Given a group like `datamart-delta-some-role-some-org` there's no perfect way of determining which bit is an organisation id vs role, they can both have hyphens in, and there's nothing stopping a role suffix matching an organisation prefix.

I've taken the approach from Delta:

* Get a list of organisation ids from MarkLogic
* Have a hardcoded list of system roles, plus a list of all access groups from AD
* Match the group suffix against an organisation id if possible, longest wins
* Attempt to match the rest of the group name against a system role or access group

Invalid roles get discarded, as do (role, org) memberships where the user isn't part of both the role (datamart-delta-role) and the organisation (datamart-delta-user-org). The next step would be to construct the SAML token based on these instead so that invalid permissions get dropped from MarkLogic, but I haven't done that yet.

Neither the organisation or access group lookup are cached, so they both happen every login. I've added logs so we can figure out how long it takes, but I think it will be just about acceptable (~1 second), and we can add caching if required.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/MemberOfToDeltaRolesMapper.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/MemberOfToDeltaRolesMapper.kt
@@ -170,4 +170,8 @@ val DELTA_SYSTEM_ROLES = hashSetOf(
     "stats-data-certifiers",
 )
 
+/*
+ * Name is copied from Delta, "External" in that it's managed by MarkLogic rather than Delta's Java layer.
+ * We should be able to get rid of this distinction at some point, but for now Delta expects it.
+ */
 val DELTA_EXTERNAL_ROLES = hashSetOf("section-151-officers")
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/MemberOfToDeltaRolesMapper.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/MemberOfToDeltaRolesMapper.kt
@@ -1,0 +1,173 @@
+package uk.gov.communities.delta.auth.services
+
+import kotlinx.serialization.Serializable
+import org.slf4j.LoggerFactory
+
+typealias MemberOfToDeltaRolesMapperFactory = (
+    username: String,
+    allOrganisationIds: List<String>,
+    allAccessGroups: List<AccessGroup>,
+) -> MemberOfToDeltaRolesMapper
+
+class MemberOfToDeltaRolesMapper(
+    private val username: String,
+    allOrganisationIds: List<String>,
+    allAccessGroups: List<AccessGroup>,
+) {
+    companion object {
+        const val DATAMART_DELTA_PREFIX = "datamart-delta-"
+        private val logger = LoggerFactory.getLogger(MemberOfToDeltaRolesMapper::class.java)
+    }
+
+    @Serializable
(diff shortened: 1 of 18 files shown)


auth-service/src/test/kotlin/uk/gov/communities/delta/security/RegistrationServiceTest.kt
```diff
@@ -48,7 +48,7 @@ class RegistrationServiceTest {
     fun testRegisteringNewStandardUser() = testSuspend {
         val registrationResult = registrationService.register(
             Registration("Test", "User", "user@example.com"),
-            listOf(Organisation(orgCode)),
+            listOf(Organisation(orgCode, name = "Test org")),
```
hugh-emerson, line 51: Are you able to skip  `name =` if the optional parameter (retirement date) is moved to the end? 

And just slightly inconsistent - omitting the optional param here vs passing in null in OAuthSSOLoginTest.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/MemberOfToDeltaRolesMapper.kt
```diff
@@ -0,0 +1,173 @@
...
+    "abatements-approvers",
+    "suspensions-approvers",
+    "write-offs-approvers",
+    "setup-managers",
+    "report-users",
+    "personal-data-owners",
+    "stats-data-certifiers",
+)
+
+val DELTA_EXTERNAL_ROLES = hashSetOf("section-151-officers")
```
hugh-emerson, line 173: I think I'm missing some context. What's makes section-151-officers an "external" role and need to be handled differently?

hugh-emerson (approved):

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/MemberOfToDeltaRolesMapper.kt
BenRamchandani: I think because it's managed by MarkLogic rather than Delta's Java.. though that's a distinction which doesn't currently matter here, could merge them and just have a comment

BenRamchandani: Just added a comment for now, Delta needs the distinction, though it doesn't do much with this role

auth-service/src/test/kotlin/uk/gov/communities/delta/service/MemberOfToDeltaRolesMapperTest.kt
```diff
@@ -0,0 +1,112 @@
...
+import kotlin.test.assertEquals
+
+class MemberOfToDeltaRolesMapperTest {
+    @Test
+    fun testOrganisationMapping() {
+        val result = mapper().map(
+            listOf(
+                "user-dclg"
+            ).map { "datamart-delta-$it" }
+        )
```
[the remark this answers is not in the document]
BenRamchandani: Done as a second test at the bottom :)

auth-service/src/test/kotlin/uk/gov/communities/delta/security/RegistrationServiceTest.kt
BenRamchandani: Makes sense, done

PR #36, opened 2023-10-09
Migration to remove dup index on set_password_tokens user_cn

BenRamchandani:
An identical index was made automatically for the primary key


hugh-emerson (approved):
I'll take your word for it

PR #37, opened 2023-10-09
DT-551: catch and handle ldap error 32 (not found) when searching for…

hugh-emerson:
… groups

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/RegistrationService.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/RegistrationService.kt
@@ -73,7 +73,7 @@ class RegistrationService(
     }
 
     private suspend fun addUserToOrganisations(adUser: UserService.ADUser, organisations: List<Organisation>) {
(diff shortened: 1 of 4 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/RegistrationService.kt
```diff
@@ -73,22 +73,23 @@ class RegistrationService(
...
+        logger.atInfo().addKeyValue("UserDN", adUser.dn).log("Adding user to domain organisations")
         try {
             organisations.forEach {
                 if (!it.retired) {
                     groupService.addUserToGroup(
                         adUser,
                         organisationUserGroup(it.code)
                     )
-                    logger.atInfo().addKeyValue("UserDN", adUser.dn)
-                        .log("Added user to domain organisation with code {}", it.code)
```
PhoebeJackson1, line 85: Do we not need to know which organisations specifically the user is added to for audit purposes?

PhoebeJackson1 (approved):
Looks good - thanks!

PR #41, opened 2023-10-10
Refactor registration error handling

BenRamchandani:
Move the email error handling into a separate try/catch block. Logic should all be the same, but I think this is a bit easier to follow


PhoebeJackson1 (approved):
Looks good to me!

PR #46, opened 2023-10-12
Fix SMTP error + add timeout

BenRamchandani:
I think what happened on prod this morning:

* Someone tried to register, and it tried to send an email
* It needed the SSL socket factory setting, it was trying to send to the TLS port over plaintext
  * We must not have tested this on staging, sorry I should have done that when testing, I just tried test
* This took a long time/possibly forever to timeout
  * If it ever did timeout Ktor's error handler logs CancellationExceptions (from the HTTP request being cancelled) at Debug level, so we wouldn't have seen it
* It's a blocking call on one of the default threads, seems like it kept getting some requests waiting for it
* So then some requests started timing out waiting to be picked up, eventually enough of them did that ECS replaced it, then the next one would work fine until someone tried to register

This PR:

* Fixes the bug, adding `props["mail.smtp.socketFactory.class"] = "javax.net.ssl.SSLSocketFactory"`
* Adds timeouts to SMTP requests
* Moves email sending to the IO thread pool
* Enables Debug logging for the "Application" logger, so cancellation exceptions get logged
* Adds more logs for how long different operations take

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
@@ -58,7 +58,7 @@ fun Route.externalRoutes(
     deltaLoginController: DeltaLoginController,
     deltaSSOLoginController: DeltaSSOLoginController,
     deltaUserRegistrationController: DeltaUserRegistrationController,
-    deltaSetPasswordController: DeltaSetPasswordController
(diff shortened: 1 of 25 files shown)


hugh-emerson:
> We must not have tested this on staging, sorry I should have done that when testing, I just tried test

Yeah I used the test environment when testing it. I guess we should always test on staging.

hugh-emerson (approved):

PR #50, opened 2023-10-17
DT-570: Testing fix

PhoebeJackson1:
Fix bug raised in testing (see comment: https://dluhcdigital.atlassian.net/browse/DT-570?focusedCommentId=441368)

> Mostly works, but if you register, load the set password page, then wait for your token to expire, you get the Send new activation email page, but pressing the button errors because the token has been deleted

All automated tests pass, manual testing of the expired token scenarios passes

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/PasswordTokenService.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/PasswordTokenService.kt
@@ -78,24 +78,24 @@ abstract class PasswordTokenService(private val tableName: String, private val d
 
     suspend fun validateToken(token: String, userCN: String): TokenResult {
         return withContext(Dispatchers.IO) {
-            readToken(token, userCN, false)
+            readToken(token, userCN)
         }
     }
 
-    suspend fun consumeToken(token: String, userCN: String): TokenResult {
+    suspend fun consumeTokenIfValid(token: String, userCN: String): TokenResult {
         return withContext(Dispatchers.IO) {
-            readToken(token, userCN, true)
+            val tokenResult = readToken(token, userCN)
+            if (tokenResult is ValidToken) {
+                deleteToken(token, userCN)
+            }
+            tokenResult
         }
     }
 
     @Blocking
-    private fun readToken(token: String, userCN: String, delete: Boolean): TokenResult {
+    private fun readToken(token: String, userCN: String): TokenResult {
         dbPool.useConnection {
-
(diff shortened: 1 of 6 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/PasswordTokenService.kt
```diff
@@ -78,24 +78,24 @@ abstract class PasswordTokenService(private val tableName: String, private val d
...
 
-    suspend fun consumeToken(token: String, userCN: String): TokenResult {
+    suspend fun consumeTokenIfValid(token: String, userCN: String): TokenResult {
         return withContext(Dispatchers.IO) {
-            readToken(token, userCN, true)
+            val tokenResult = readToken(token, userCN)
+            if (tokenResult is ValidToken) {
+                deleteToken(token, userCN)
+            }
+            tokenResult
```
BenRamchandani, line 91: This could let two concurrent requests both consume the token, which doesn't really matter too much, but we should try and prevent it. `deleteToken` should check `executeUpdate` returns 1 to indicate a row was deleted.

Ideally the read would be a `SELECT FOR UPDATE` if we're consuming it to lock the row, and then the delete happens in the same transaction (so in one `useConnection` block), but that's a bit more faff.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/PasswordTokenService.kt
```diff
@@ -78,23 +78,27 @@ abstract class PasswordTokenService(private val tableName: String, private val d
...
-    suspend fun consumeToken(token: String, userCN: String): TokenResult {
+    suspend fun consumeTokenIfValid(token: String, userCN: String): TokenResult {
         return withContext(Dispatchers.IO) {
-            readToken(token, userCN, true)
+            var tokenResult = readToken(token, userCN)
+            if (tokenResult is ValidToken) {
+                val tokenDeletedSuccessfully = deleteToken(token, userCN)
+                if (!tokenDeletedSuccessfully)
+                    tokenResult = NoSuchToken
+                logger.warn("Token not deleted successfully")
```
BenRamchandani, line 92:
```suggestion
                if (!tokenDeletedSuccessfully) {
                    tokenResult = NoSuchToken
                    logger.warn("Token not deleted successfully")
                }
```

BenRamchandani (approved):
Missing braces, but otherwise looks good

PR #52, opened 2023-10-24
DT-569: small fixes from testing

hugh-emerson:
Fixes two small issues found in testing:
* Email title appears as `${title}` in the email content in some mail apps
* Email address doesn't appear in confirmation message because of var name typo

Also here are some other observations from testing that I haven't addressed because they're not super important:

* "Reset Password Error" - shouldn't this error appear when you land on the page rather than after you submit the new password? I know the validation has to be done on submit, so probably not worth also validating on page load just to improve UX for a rare scenario
* Request IDs - I think we might be too eager to tell the user to contact support with a request ID. It makes the error message a bit busier/intimidating and we should have a bit more faith in the app - if the system says their reset token is invalid, then it's invalid for a good reason and they should not contact support.

Also, I hit the rate limit on the site without triggering any CloudWatch alarm. Is that expected?

### DIFF
--- a/auth-service/src/main/resources/templates/thymeleaf/emails/fragments/template.html
+++ b/auth-service/src/main/resources/templates/thymeleaf/emails/fragments/template.html
@@ -3,7 +3,7 @@
 
 <head th:fragment="email-head (title)">
     <meta charset="utf-8"/>
-    <title>${title}</title>
+    <title th:text="${title}"></title>
     <style>
         p {
             font-family: calibri, serif;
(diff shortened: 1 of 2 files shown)


PhoebeJackson1 (approved):

PhoebeJackson1:
When you say "Reset Password Error" what do you mean? I think the checking of the token should happen on get request and post requests so it should show as soon as it is aware of an issue?

The point about request IDs makes sense, I will make sure to consider where to use them more in future :) 

It should trigger an alarm for the rate limit - I'll investigate this now

PR #54, opened 2023-11-23
DT-612 Audit log

BenRamchandani:
Start adding an audit log, currently only auditing logins.

See [DT-612](https://dluhcdigital.atlassian.net/browse/DT-612) for a HLD.

I think the service (business logic, suspend functions) and repositories (plumbing, in this case database queries that are blocking) pattern fits quite well here. It's nice to divide blocking + suspend, and it should make it easy to combine queries into larger transactions which isn't straightforward at the moment. If everyone else is happy with it I'll do some refactoring of the other services too at some point.

[DT-612]: https://dluhcdigital.atlassian.net/browse/DT-612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
@@ -167,6 +167,7 @@ class Injection(
             authorizationCodeService,
             failedLoginCounter,
             successfulLoginCounter,
+            userAuditService,
         )
     }
 
@@ -241,6 +242,7 @@ class Injection(
             registrationService,
             organisationService,
             ssoLoginCounter,
+            userAuditService,
         )
 
     fun groupService() = GroupService(ldapService, ldapConfig)
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
@@ -130,6 +130,7 @@ class Injection(
     val failedLoginCounter: Counter = meterRegistry.counter("login.failedLogins")
     val loginRateLimitCounter: Counter = meterRegistry.counter("login.rateLimitedRequests")
     val successfulLoginCounter: Counter = meterRegistry.counter("login.successfulLogins")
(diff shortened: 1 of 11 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
```diff
@@ -128,6 +130,7 @@ class Injection(
     val failedLoginCounter: Counter = meterRegistry.counter("login.failedLogins")
     val loginRateLimitCounter: Counter = meterRegistry.counter("login.rateLimitedRequests")
     val successfulLoginCounter: Counter = meterRegistry.counter("login.successfulLogins")
+    val ssoLoginCounter: Counter = meterRegistry.counter("login.ssoLogins")
```
PhoebeJackson1, line 133: Does this need an alarm added to it like the others? Or is it just for reference?

auth-service/src/test/kotlin/uk/gov/communities/delta/security/OAuthSSOLoginTest.kt
```diff
@@ -324,14 +333,17 @@ class OAuthSSOLoginTest {
             ldapUserLookupServiceMock = mockk<UserLookupService>()
             registrationService = mockk<RegistrationService>()
             organisationService = mockk<OrganisationService>()
+            userAuditService = mockk<UserAuditService>()
+            loginCounter = mockk<Counter>()
```
PhoebeJackson1, line 337: Could this be called SSO login counter just to make it obvious that it is different from the "successfulLoginNoopCounter" counter used for the loginPageController?

PhoebeJackson1 (approved):
:)

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Injection.kt
BenRamchandani: I think just for reference and to add to a dashboard at some point, it's Microsoft's job to worry about rate limiting for this.

PR #60, opened 2023-11-27
Use jakarta email validation

hugh-emerson:


BenRamchandani (approved):
Looks good. Sensible anyway, but I'm curious what was the email that caused this? I thought Delta has sent emails to "delta.admin" before and had it work (to mailhog)

PR #61, opened 2023-11-30
DT-612 Add CSV audit log export

BenRamchandani:
And (very) basic pagination logic that I'll flesh out when doing the Delta side changes.

I've made the CSV headers dynamic based on the top level attributes of "actionData". This feels slightly crazy, but I can't think of a better way of doing it, returning JSON as a CSV column sounds ugly, and whilst at the moment there's nothing important in there, when we're auditing user changes there will be. I think it would only really be an issue if you wanted to combine or compare audit logs for different users, but I don't see why you would.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/FetchUserAuditController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/FetchUserAuditController.kt
@@ -35,7 +35,7 @@ class FetchUserAuditController(
         val session = principal<OAuthSession>()!!
         val page = parameters["page"]?.toInt()
         val audit = getUserAudit(session, parameters.getOrFail("cn"), page)
-        return respond(audit.map {
+        return respond(mapOf("userAudit" to audit.map {
             JsonObject(
                 mapOf(
                     "action" to JsonPrimitive(it.action.action),
@@ -46,7 +46,7 @@ class FetchUserAuditController(
                     "actionData" to it.actionData,
                 )
             )
-        })
+        }))
     }
 
     private suspend fun ApplicationCall.getUserAuditCSV() {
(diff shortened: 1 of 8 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/StatusPages.kt
```diff
@@ -160,7 +171,19 @@ private suspend fun ApplicationCall.respondStatusPage(statusError: StatusErrorPa
...
+        "error_description" to "User visible error ${statusError.userErrorPageTitle}",
+        "user_visible_error" to statusError.userErrorPageMessage,
+    ))
+}
+
+private suspend fun ApplicationCall.apiErrorResponse(apiError: ApiError) {
+    respond(apiError.statusCode, mapOf(
+        "error" to apiError.errorCode,
+        "error_description" to apiError.errorDescription,
+        "user_visible_error" to apiError.userVisibleMessage,
```
BenRamchandani, line 185: The idea is this can then be shown to the user by Delta

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/utils/CSVUtil.kt
```diff
@@ -0,0 +1,19 @@
...
+            append('"')
+            append(element.replace("\"", "\"\""))
+            append('"')
+            if (i < row.size - 1) append(',')
+        } else {
+            append(element)
+            if (i < row.size - 1) append(',')
+        }
+    }
+    append('\n')
```
BenRamchandani, line 18: I looked for a library for this, but they mostly take objects and I've added dynamic columns for the actionData attributes, then it didn't seem worth pulling in a library for a 20 line helper function

PhoebeJackson1 (approved):

PhoebeJackson1 (commented):
(GitHub only showed me 2 files the first time and I didn't notice there were more oops sorry!)

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/FetchUserAuditController.kt
```diff
@@ -10,34 +10,32 @@ import kotlinx.serialization.json.JsonObject
...
     private val userLookupService: UserLookupService,
     private val userAuditService: UserAuditService,
 ) {
     private val logger = LoggerFactory.getLogger(javaClass)
+    private val pageSize = 100
 
     fun route(route: Route) {
-        route.get { getUserAudit(call) }
+        route.get { call.getUserAuditJson() }
+        route.get("/csv") { call.getUserAuditCSV() }
```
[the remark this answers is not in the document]
BenRamchandani: It should only match the exact route rather than being a catch-all

PR #63, opened 2023-12-06
DT-572 Display message on register page for non-required SSO

BenRamchandani:
A nice to have that we skipped before.

When a user signs in with SSO, and they don't have an associated Delta account, and SSO is not required for their domain (so they don't automatically have an account created) they get redirected to the register page. This change displays a message, and pre-fills the email inputs.

register page screenshot

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaUserRegistrationController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaUserRegistrationController.kt
@@ -48,6 +48,15 @@ class DeltaUserRegistrationController(
     }
 
     private suspend fun registerGet(call: ApplicationCall) {
+        val ssoEmail = call.parameters["sso_email"]
+        if (!ssoEmail.isNullOrEmpty()) {
+            call.respondRegisterPage(
+                emailAddress = ssoEmail,
+                confirmEmailAddress = ssoEmail,
+                otherErrors = listOf("No Delta account found with email $ssoEmail. Please register below, or contact the service desk if you already have an account with a different email."),
+                errorSummary = "Account not found",
+            )
+        }
         call.respondRegisterPage()
     }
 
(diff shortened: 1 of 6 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaUserRegistrationController.kt
```diff
@@ -48,6 +48,15 @@ class DeltaUserRegistrationController(
     }
 
     private suspend fun registerGet(call: ApplicationCall) {
+        val ssoEmail = call.parameters["sso_email"]
```
PhoebeJackson1, line 51: Could this variable be named something else, I'm not sure quite what best in my head is "notRequiredSSOEmail" which is rather gross but I think just ssoEmail is a little confusing to read? Maybe "fromSSOEmail"?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/StatusPages.kt
```diff
@@ -29,7 +29,8 @@ fun Application.configureStatusPages(deltaWebsiteUrl: String, ssoConfig: AzureAD
                             "register-user-form",
                             mapOf(
                                 "deltaUrl" to deltaConfig.deltaWebsiteUrl,
-                                "allErrors" to arrayListOf(arrayListOf(tooManyRequestsErrorMessage, "#")),
+                                "allErrors" to arrayListOf(Pair(tooManyRequestsErrorMessage, "#")),
+                                "errorSummary" to "Rate limit reached",
```
PhoebeJackson1, line 33: Does this still fit the gov guidance to have text other than "There is a problem" here? I agree it's clearer for sure tho

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/DeltaUserRegistrationControllerTest.kt
```diff
@@ -32,6 +32,14 @@ class DeltaUserRegistrationControllerTest {
         }
     }
 
+    @Test
+    fun testRegisterPageFromSSO() = testSuspend {
+        testClient.get("/register?sso_email=user@example.com").apply {
```
PhoebeJackson1, line 37: Should this be encoded? Or is that not necessary?

PhoebeJackson1 (approved):
Best error handling on delta 😂

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/plugins/StatusPages.kt
BenRamchandani: I guess not, it does say it should always be "There is a problem". I think it's clearer anyway so I'll leave it

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/DeltaUserRegistrationControllerTest.kt
BenRamchandani: It does work, but it probably should be

PR #66, opened 2023-12-15
DT 610 - Emails through auth service

PhoebeJackson1:
Set up delta to send the activation and reset password emails triggered by admins through the auth service. Did some refactoring of emails related code. Added tests.

Delta PR to follow shortly.
Splitting the work for the ticket into 2 PRs so this is only the first step in the HLD from the ticket.

Testing: Lots done locally and tests written
Risks: If emails don't send correctly then it could cause an increase in support issues and/or confusion
Documentation: Ticket and this PR (no major changes to the user)

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/EmailService.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/EmailService.kt
@@ -58,12 +58,12 @@ class EmailService(
         logger.atInfo().addKeyValue("userCN", userCN).log("Sent already-a-user email")
     }
 
-    suspend fun sendSetPasswordEmail(user: LdapUser, token: String, triggeredByAdmin: Boolean, call: ApplicationCall) {
+    suspend fun sendSetPasswordEmail(user: LdapUser, token: String, triggeringAdminSession: OAuthSession?, call: ApplicationCall) {
         sendSetPasswordEmail(
             user.firstName,
             token,
             user.cn,
-            triggeredByAdmin,
+            triggeringAdminSession,
             EmailContacts(user.email!!, user.fullName, emailConfig),
             call
         )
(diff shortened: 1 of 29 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserAuditService.kt
```diff
@@ -41,7 +41,7 @@ class UserAuditService(private val userAuditTrailRepo: UserAuditTrailRepo, priva
         )
     }
 
-    val userForgotPasswordAudit = insertSimpleAuditRowFun(UserAuditTrailRepo.AuditAction.FORGOT_PASSWORD_EMAIL)
+    val resetPasswordEmailAudit = insertSimpleAuditRowFun(UserAuditTrailRepo.AuditAction.RESET_PASSWORD_EMAIL)
```
BenRamchandani, line 44: We should record the calling user in the `editing_user_cn` column of the table. leaving it null only if the user triggered it themselves.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/EmailService.kt
```diff
@@ -177,11 +185,17 @@ class EmailService(
         logger.atInfo().addKeyValue("userCN", userCN).log("Sent password-never-set email")
     }
 
-    suspend fun sendResetPasswordEmail(user: LdapUser, token: String, call: ApplicationCall) {
+    suspend fun sendResetPasswordEmail(
+        user: LdapUser,
+        token: String,
+        triggeredByAdmin: Boolean,
```
BenRamchandani, line 191: nitpick, don't need to change it, but would be nicer to have it be e.g. `triggeringAdminSession: OAuthSession?` or similar. It's a bit more faff, but it makes it much harder to pass the wrong thing in, means sending an email doesn't need to know about the principal in the application call, and is clearer to read (hard to know what true means in `sendResetPasswordEmail(user, token, true, call)` without reading the function. Generally passing optional thing in > enum > boolean flag).

BenRamchandani (approved):

BenRamchandani (approved):

PR #67, opened 2023-12-15
DT-744 Redirect user to error page if old timestamp passed to login

BenRamchandani:
This currently redirects to Delta's "Sign-in page expired" page. An alternative would be to redirect to `/oauth2/authorization/delta-auth` and have them bounce straight back to the login page, not sure what's better.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
@@ -45,18 +45,22 @@ class DeltaLoginController(
         }
     }
 
+    // If the user lands on the login page with a stale state parameter we redirect them back to Delta to get a fresh one
+    private val loginPageValidDuration = 1.hours
+
     private suspend fun loginGet(call: ApplicationCall) {
         val params = call.getLoginQueryParams()
         if (params == null) {
             logger.info("Invalid parameters for login request, redirecting back to Delta")
             return call.respondRedirect(deltaConfig.deltaWebsiteUrl + "/login?error=delta_invalid_params&trace=${call.callId!!.encodeURLParameter()}")
         }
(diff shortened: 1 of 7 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
```diff
@@ -50,6 +51,11 @@ class DeltaLoginController(
             logger.info("Invalid parameters for login request, redirecting back to Delta")
             return call.respondRedirect(deltaConfig.deltaWebsiteUrl + "/login?error=delta_invalid_params&trace=${call.callId!!.encodeURLParameter()}")
         }
+        if (params.timestamp?.let { it < (System.currentTimeMillis() / 1000) - 1.hours.inWholeSeconds } == true) {
```
PhoebeJackson1, line 54: Could you make a const for the expiry time so that if it gets changed it is changed here and wherever it is checked on post requests?

PhoebeJackson1 (approved):

PhoebeJackson1:
I would say that not having to do an extra click as a user is better 🤷‍♀️ so straight back to the login page and they never know what happened?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
BenRamchandani: It's Delta that checks it when the user is redirected back after logging in. TBH I don't know how long they're even valid for, it's checked by Spring Security somewhere.

I'll pull out a constant anyway

BenRamchandani:
> I would say that not having to do an extra click as a user is better 🤷‍♀️ so straight back to the login page and they never know what happened?

Yeah makes sense, I've updated it 🙂

PhoebeJackson1 (approved):

PR #68, opened 2023-12-15
Give error message on origin header check

BenRamchandani:
This should nudge users away from Internet Explorer without them having to raise a ticket, and mean it stops appearing in the error logs.

We get someone trying to use IE every couple of weeks, but it doesn't seem to support the `Origin` header which we use for CSRF protection.


hugh-emerson (approved):
Seems sensible

PR #71, opened 2023-12-18
Update alarm descriptions

BenRamchandani:


terraform/modules/auth_service/monitoring.tf
```diff
@@ -37,7 +42,12 @@ resource "aws_cloudwatch_metric_alarm" "success_rate_high_login" {
   threshold           = 30 // Highest in last month (10/08) is 23
   treat_missing_data  = "notBreaching"
 
-  alarm_description = "There are more successful logins than expected"
+  alarm_description = <<EOF
+There are more failed logins than expected.
```
PhoebeJackson1, line 46: *successful

terraform/modules/auth_service/monitoring.tf
```diff
@@ -117,7 +143,11 @@ resource "aws_cloudwatch_metric_alarm" "auth_forgot_password_rate_limit_reached"
   threshold           = 1
   treat_missing_data  = "notBreaching"
 
-  alarm_description = "An IP address has reached the rate limit on the forgot password page"
+  alarm_description = <<EOF
+An IP address has reached the rate limit on the reset password page
```
PhoebeJackson1, line 147: *forgot password

terraform/modules/fargate/monitoring.tf
```diff
@@ -80,7 +80,11 @@ resource "aws_cloudwatch_metric_alarm" "healthy_host_low_alb" {
   statistic           = "Minimum"
   threshold           = var.desired_count
 
-  alarm_description  = "There are fewer healthy hosts than expected"
+  alarm_description  = <<EOF
+There are fewer healthy ECS tasks than expected in the ${aws_ecs_cluster.main.name} cluster.
```
PhoebeJackson1, line 84: Does healthy hosts = healthy tasks?

BenRamchandani: Yeah, I normally say "hosts" to mean EC2 instances and "tasks" to mean ECS tasks, but "host" is more general really.

This alarm is based on the load balancer health checking its targets and it doesn't know what those targets are, HealthyHostCount is a count of how many healthy targets it has.

BenRamchandani: Slightly confused by the fact that ECS can do it's own healtchecks that we don't use

PhoebeJackson1 (approved):

PR #78, opened 2024-01-24
Dt 610 2

PhoebeJackson1:
Pull request 2 of 3 for this ticket 🤞 
From HLD: 
2. Create User:
- Create endpoint in auth service and send request to it upon clicking "Save" button
- Authenticate user triggering the request as above
- Create user as with registration (with more fields)
- Add audit logs
- Send activation email to relevant user (and audit log this)
- Handle errors or success and communicate those back to delta

Delta passes the already validated user details to the auth service. Auth service then validates the request (but not the data) and makes the request(s) to AD to make the user. Auth service audits this user creation. Auth service sends activation email to user (thus delta will no longer send any activation emails). Auth service responds to Delta or sends a (hopefully better than before) error message which delta displays. Delta repo PR to come soon.

Testing: I have testing this lots locally (using test ad) and have added tests to the code as well
Risks: There are a few nichenesses to the delta code and places where it handles data that in all my tests have been blank (e.g. external roles) - I have investigated any of these that I have found and believe they are not relevant but it is possible I missed something, in this case I think the outcome would be that a user would be unable to do something they expected to at which point they would hopefully raise a support ticket allowing us to know about the issue, fix it (manually or not) and correct the code -> only a very small risk of major impact
Documentation: Ticket and this PR (and eventually the delta PR), if there is any other place I should document this please let me know!)

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
@@ -1,6 +1,11 @@
 package uk.gov.communities.delta.auth.services
 
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.ktor.server.application.*
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
 import org.slf4j.LoggerFactory
+import uk.gov.communities.delta.auth.config.AzureADSSOClient
 import uk.gov.communities.delta.auth.config.LDAPConfig
 import uk.gov.communities.delta.auth.controllers.external.ResetPasswordException
 import uk.gov.communities.delta.auth.utils.randomBase64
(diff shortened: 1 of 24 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
```diff
@@ -10,31 +15,101 @@ import javax.naming.directory.*
...
+        val auditData = mutableMapOf<String, String>()
+        auditData["userPrincipalName"] = adUser.userPrincipalName
+        auditData["cn"] = adUser.cn
+        auditData["sn"] = adUser.sn
+        auditData["givenName"] = adUser.givenName
+        auditData["mail"] = adUser.mail
+        auditData["st"] = adUser.notificationStatus
+        auditData["userAccountControl"] = adUser.userAccountControl
+
+        if (adUser.comment != null) auditData["comment"] = adUser.comment!!
```
BenRamchandani, line 78: It's annoying we have to use `!!`. Could do `adUser.comment?.let { auditData["comment"] = it }` but up to you whether you think that's clearer

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
```diff
@@ -124,6 +259,10 @@ class UserService(
             return String.format("%s@%s", cn, ldapConfig.domainRealm)
         }
 
+        fun getDisplayName(): String {
+            return this.givenName + this.sn
```
BenRamchandani, line 263: Should this have a space?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
```diff
@@ -148,4 +287,24 @@ class UserService(
             }
         }
     }
+
+    data class DeltaUserDetails(
+        @JsonProperty("id") val id: String, //Not used anywhere yet
```
BenRamchandani, line 292: What is id, is it the user's CN?

PhoebeJackson1: It comes through as blank on the POST request, I am assuming it will become clear when editing users, I will add a TODO to make sure I know what it does (and add a comment/clarify it) by the end of the next part of the ticket if it's ok to be left for now?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
PhoebeJackson1: Oopsie thanks!

BenRamchandani (approved):

BenRamchandani (approved):

PR #79, opened 2024-01-29
Allow empty password for SSO users

BenRamchandani:
Mostly useful on staging since there's no button for the Softwire SSO.


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
```diff
@@ -151,6 +145,13 @@ class DeltaLoginController(
             return call.respondRedirect(oauthClientLoginRoute(ssoClientMatchingEmailDomain.internalId, formUsername))
         }
 
+        if (password.isNullOrEmpty()) return call.respondLoginPage(
+            client,
+            errorMessage = "Password is required",
+            errorLink = "#password",
+            username = formUsername,
+        )
```
hugh-emerson, line 153: Should tests be updated to cover this case? E.g. update testLoginPostSSODomainRedirects to not include a password?

hugh-emerson (approved):

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/DeltaLoginController.kt
BenRamchandani: Good idea

PR #80, opened 2024-02-06
DT-585: use the full postgres CA cert bundle, not just one

hugh-emerson:
`openssl x509` only uses the first cert in the file. 

I've manually deployed to test and updated the test SQL cert. Works fine. I wanted to do it with an env var defined in the Dockerfile originally (so that the path is only referenced in one place) but that didn't seem to have any effect.

Will do the same for keycloak in the delta repo.


BenRamchandani (approved):
Nice :+1:

PR #81, opened 2024-02-14
DT 610 3

PhoebeJackson1:
Use the Auth service for the update user changes by an admin

Included assessment of what admin roles should be allowed to do what actions, decided on:
Sending emails:
    Read only admin
    admin
viewing user
    read only admin
    admin
editing user
    admin
creating user
    admin
enabling/disabling user (not implemented in this ticket)
    read only admin
    admin
resetting mfa (not implemented in this ticket)
    admin
changing notification setting (not implemented in this ticket)
    read only admin
    admin
    
Also identified that currently users can be enabled while not having passwords so disabled all accounts in this state on prod, staging and test.

Testing: Lots locally and will add more to the ticket for once merged
Risks: Hopefully movement of security checks to the auth service will make them more secure, it is possible an edge case has been missed but since the admin user can see the user after the changes hopefully the effects of this risk are minimal 
Documentation: This PR, the ticket and the PR on delta (to follow)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/GroupService.kt
```diff
@@ -45,19 +55,50 @@ class GroupService(
...
+        userDN: String,
+        groupCN: String,
+        call: ApplicationCall,
+        triggeringAdminSession: OAuthSession? = null,
+    ) {
+        val groupDN = ldapConfig.groupDnFormat.format(groupCN)
+
+        if (!groupExists(groupDN)) {
+            throw Exception("Group did not exist in AD when trying to remove a user from it")
+        }
```
BenRamchandani, line 94: Do we not get a nice error message without this?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/GroupService.kt
```diff
@@ -45,19 +55,50 @@ class GroupService(
...
+    ) {
+        addUserToGroup(adUser.cn, adUser.dn, groupCN, call, triggeringAdminSession)
+    }
+
+    suspend fun addUserToGroup(
+        userCN: String,
+        userDN: String,
+        groupCN: String,
+        call: ApplicationCall,
+        triggeringAdminSession: OAuthSession? = null,
```
BenRamchandani, line 67: I think I'd rather this wasn't optional, makes it harder to leave it out by accident

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditUserController.kt
```diff
@@ -0,0 +1,154 @@
...
+
+    private fun getModifications(
+        currentUser: LdapUser,
+        newUser: UserService.DeltaUserDetails
+    ): Array<ModificationItem> {
+        var modifications = arrayOf<ModificationItem>()
+
+        getModificationItem("sn", currentUser.lastName, newUser.lastName)?.let { modifications += it }
+        getModificationItem("givenName", currentUser.firstName, newUser.firstName)?.let { modifications += it }
+        getModificationItem("mail", currentUser.email, newUser.email)?.let { modifications += it }
```
BenRamchandani, line 124: We shouldn't allow changing this, it has to stay in sync with the cn

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/GroupService.kt
PhoebeJackson1: Its a NameNotFoundException which aren't the easiest to decipher but since they do actually have the information and this adds slowness I have removed these lines :)

BenRamchandani (approved):

PhoebeJackson1:
Raised ticket https://dluhcdigital.atlassian.net/browse/DT-896 to move the remaining buttons on this page

BenRamchandani (approved):

PR #83, opened 2024-02-16
DT-277 Map of old to new user objectGUIDs as table

BenRamchandani:
Adds a new overnight job to import all user old and actual objectGUIDs into a table in the auth service.

I'm planning to export it as a CSV and load it into MarkLogic to use for updating the objectGUIDs there. We're not doing anything with the map in the auth service, it's mostly a convenient place to add the script.

The job takes a few seconds on test. It will be longer on prod, but I think <30 seconds still. It loads the full list of users into memory twice, but napkin maths says it should be OK.

One annoying thing is that Active Directory renders GUIDs to strings slightly differently to Java given the same bytes. I'm planning to use the Java representation for the strings in MarkLogic since we're expecting to move away from AD anyway long term.

### DIFF
--- a/auth-service/build.gradle.kts
+++ b/auth-service/build.gradle.kts
@@ -85,7 +85,7 @@ dependencies {
 // Migrations are run by the application on startup, or on first use of the database in Development mode.
 // Extra task to run Flyway migrations without (re)starting the app.
 task("migrate", JavaExec::class) {
-    mainClass.set("uk.gov.communities.delta.auth.services.DbPoolKt")
+    mainClass.set("uk.gov.communities.delta.auth.repositories.DbPoolKt")
     classpath = sourceSets["main"].runtimeClasspath
 }
 
--- a/auth-service/entrypoint.sh
+++ b/auth-service/entrypoint.sh
@@ -7,18 +7,18 @@ function echo_json() {
 
 echo_json "Entrypoint script starting"
 
-if [ -v RUN_TASK ]; then
-    echo_json "RUN_TASK is set, starting task ${RUN_TASK}"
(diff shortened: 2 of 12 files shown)


auth-service/build.gradle.kts
```diff
@@ -85,7 +85,7 @@ dependencies {
 // Migrations are run by the application on startup, or on first use of the database in Development mode.
 // Extra task to run Flyway migrations without (re)starting the app.
 task("migrate", JavaExec::class) {
-    mainClass.set("uk.gov.communities.delta.auth.services.DbPoolKt")
+    mainClass.set("uk.gov.communities.delta.auth.repositories.DbPoolKt")
```
BenRamchandani, line 88: Unrelated fix, had the wrong package

auth-service/entrypoint.sh
```diff
@@ -7,20 +7,20 @@ function echo_json() {
...
   openssl x509 -inform der -in /tmp/dluhcldapsca.crt -outform pem -out /tmp/dluhcldapsca.pem
   keytool -import -cacerts -alias dluhcldapsca -file /tmp/dluhcldapsca.pem -noprompt -storepass changeit
 fi
 
+if [ -v RUN_TASK ]; then
+    echo_json "RUN_TASK is set, starting task ${RUN_TASK}"
+    java -cp auth-all.jar uk.gov.communities.delta.auth.tasks.RunTaskMainKt
+    echo_json "Task complete, exiting"
+    exit 0
+fi
```
BenRamchandani, line 22: None of the tasks needed to talk to AD before

hugh-emerson (approved):
👍

PR #84, opened 2024-02-20
Readme update

BenRamchandani:


auth-service/README.md
```diff
@@ -30,6 +26,52 @@ faster restarts, reloading of templates).
...
+```
+
+or in PowerShell
+
+```powershell
+.\gradlew.bat run
+```
+
+Note that Gradle will still say "EXECUTING", but the logs say "Application started" the app will be running on port 8088.
+The `io.ktor.development` is automatically set when using gradle run.
```
hugh-emerson, line 42:
```suggestion
The `io.ktor.development` property is automatically set when using gradle run.
```

auth-service/README.md
```diff
@@ -12,11 +12,7 @@
```
hugh-emerson, line 12: Should line 10 recommend `docker compose up -d postgres`?

hugh-emerson (approved):

auth-service/README.md
BenRamchandani: The app in docker compose is has a profile set, so `docker compose up` will not start it, it would have to be `docker compose --profile app up`. Could change it just for clarity, but it should work as-is.

PR #85, opened 2024-02-20
Updated GOV.UK logo

keithkennedy:
Amended README.md to be more clear about updated GOV.UK frontend


BenRamchandani (approved):

BenRamchandani:
Actually shouldn't this include at least a favicon update? Have you definitely got all the image files?

keithkennedy:
> Actually shouldn't this include at least a favicon update? Have you definitely got all the image files?

Yes it should, I have added the asset images.

PR #87, opened 2024-02-21
DT-571 Retrieve and store sso status for purpose of multi-factor authentication

JamesMooreDLUHC:
There is a corresponding PR in the Delta code base.

Retrieve, store and return the sso status when logging in for the purpose of controlling whether MFA is required on login.


hugh-emerson (changes requested):
Tests need updating (currently failing)

auth-service/src/test/kotlin/uk/gov/communities/delta/dbintegration/AuthorizationCodeServiceTest.kt
```diff
@@ -20,14 +20,14 @@ class AuthorizationCodeServiceTest {
...
-        val code = service.generateAndStore("some.user", client, "traceId")
+        val code = service.generateAndStore("some.user", client, "traceId", false)
         val result = service.lookupAndInvalidate(code.code, testServiceClient("wrong-client"))
         assertNull(result)
     }
 
     @Test
     fun testRetrieveValidCode() = testSuspend {
-        val code = service.generateAndStore("some.user", client, "traceId")
+        val code = service.generateAndStore("some.user", client, "traceId", false)
```
hugh-emerson, line 30: You could add a corresponding assertion that result.isSso looks right?

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/DeltaLoginControllerTest.kt
```diff
@@ -222,8 +222,8 @@ class DeltaLoginControllerTest {
         every { failedLoginCounter.increment(1.0) } returns Unit
         every { successfulLoginCounter.increment(1.0) } returns Unit
         coEvery { userAuditService.userFormLoginAudit(any(), any()) } returns Unit
-        coEvery { authorizationCodeService.generateAndStore(any(), any(), any()) } answers {
-            AuthCode("test-auth-code", "user", client, Instant.now(), "trace")
+        coEvery { authorizationCodeService.generateAndStore(any(), any(), any(), any()) } answers {
```
hugh-emerson, line 225: Ideally, we'd have a verify somewhere that generateAndStore was called with isSso is false. But looks like it's currently not verified at all...

So I'd say, (if it's easy), could you add a suitable verify to maybe the testLoginPostSuccess test?

hugh-emerson (approved):
Very small suggestions for the tests, but otherwise looks good to me.

PR #88, opened 2024-02-22
DT-277 Use AD string representation of GUID

BenRamchandani:
Looking at your comment in AdUser.java in Delta @hugh-emerson, I think you're right that the byte representation is different and that using ADs string representation gets a "correct" UUID in that the version number is in the right place.

I couldn't find any super clear documentation on this, and I don't think it matters too much, but may as well switch.


hugh-emerson (approved):
🤷‍♂️ I can't really remember but sounds fine to me

PR #92, opened 2024-02-28
DT-610: validate email address during admin creation

hugh-emerson:


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminUserCreationController.kt
```diff
@@ -56,6 +57,16 @@ class AdminUserCreationController(
                 "A user with this username already exists, you can edit that user via the users page"
             )
         }
+        if (!emailAddressChecker.hasValidFormat(adUser.mail)) {
+            logger.atWarn().addKeyValue("email", adUser.mail).addKeyValue("UserDN", adUser.dn)
+                .log("Admin tried to create user with invalid email")
+            throw ApiError(
+                HttpStatusCode.Conflict,
```
OliviaKRose, line 64: Would a BadRequest be more appropriate here?

OliviaKRose (approved):

PR #95, opened 2024-03-05
DT-745 Impersonating user's permissions as admin

OliviaKRose:
Added new endpoint to allow admin users to impersonate users/gain their permissions
Added column to delta_session to record impersonated_users_cn for the session
Added to the impersonator's audit logs when a user is impersonated and by who

Tests:
Manually tested
Automated tests in progress


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/RefreshUserInfoController.kt
```diff
@@ -41,10 +42,50 @@ class RefreshUserInfoController(
...
+                "forbidden",
+                "User is not an enabled admin",
+                "You do not have the necessary permissions to do this"
+            )
+        }
+        val userToImpersonate = userLookupService.lookupUserByCn(impersonatedUsersCn)
+        val originalUserWithImpersonatedRoles = originalUser.copy(
+            memberOfCNs = userToImpersonate.memberOfCNs,
+            firstName = "Impersonating " + userToImpersonate.firstName,
+            lastName = userToImpersonate.lastName
```
BenRamchandani, line 74: Worth checking first name and last name don't get saved in any audit logs or anything by Delta. Or we could change the user display in Delta instead of passing a different name in here - we'd only have the email address (cn), but that might be neater/less risky.

OliviaKRose: Changed it to use the cn/email address instead. Didn't seem to be saved to the audit logs but didn't check to thoroughly as it's still probably better to use the cn

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/OAuthSessionService.kt
```diff
@@ -85,17 +86,30 @@ class OAuthSessionService(private val dbPool: DbPool, private val timeSource: Ti
...
                 traceId = authCode.traceId,
-                isSso = authCode.isSso
+                isSso = authCode.isSso,
+                impersonatedUserCn = null,
             )
         }
     }
 
+    @Blocking
+    fun updateWithImpersonatedCn(sessionId: Int, impersonatedUserCn: String) {
```
BenRamchandani, line 96: It would be nice to have a test for this too

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/RefreshUserInfoControllerTest.kt
```diff
@@ -92,22 +144,62 @@ class RefreshUserInfoControllerTest {
...
                 )
             } answers { "SAML Token" }
+            every {
+                samlTokenService.generate(
+                    client.samlCredential,
+                    adminUser,
+                    adminSession.createdAt,
+                    any()
+                )
+            } answers { "Admin SAML Token" }
```
BenRamchandani, line 154: I think this is unused

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/RefreshUserInfoControllerTest.kt
```diff
@@ -92,22 +144,62 @@ class RefreshUserInfoControllerTest {
...
+                )
+            } answers { adminSession }
+            coEvery {
+                oauthSessionService.updateWithImpersonatedCn(
+                    adminSession.id,
+                    userToImpersonate.cn,
+                )
+            } just runs
             coEvery { adminEmailController.route(any()) } just runs
+            coEvery { dbPool.useConnectionNonBlocking("impersonate_user") {} } just runs
```
BenRamchandani, line 182: Unused

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/RefreshUserInfoControllerTest.kt
```diff
@@ -64,14 +67,57 @@ class RefreshUserInfoControllerTest {
...
+        }
+    }
+
     companion object {
         private lateinit var testApp: TestApplication
         private lateinit var testClient: HttpClient
         private lateinit var controller: RefreshUserInfoController
 
         private val client = testServiceClient()
+        private val call = mockk<ApplicationCall>()
```
BenRamchandani, line 108: unused I think

BenRamchandani (approved):
Looks good. There's one more test it would be nice to have for the database update, but let's merge this so it's not blocking the Delta side and do that in another PR.

PR #97, opened 2024-03-05
Add LDAPS cert instructions to README in this repo instead of Delta

hugh-emerson:


BenRamchandani (approved):
Should put the host file bit in here, then good to merge

PR #101, opened 2024-03-08
DT-920 Endpoint for enabling and disabling users as admin

BenRamchandani:


auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/AdminEditUserControllerTest.kt
```diff
@@ -377,6 +372,7 @@ class AdminEditUserControllerTest {
                             controller,
                             mockk(relaxed = true),
                             mockk(relaxed = true),
+                            mockk(relaxed = true),
```
hugh-emerson, line 375: Really minor, but slightly annoying we have to do this. 

Is it any neater to give the bearerTokenRoutes function default values of mockk(relaxed = true) for all the parameters? Not great to have something that's purely for testing in the real app code, but it would be nicer to just do bearerTokenRoutes(adminEditUserController = controller)

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEnableDisableUserController.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        val user = findUserFromCallParameter(call)
+
+        if (user.accountEnabled) {
+            return call.respond(mapOf("message" to "Account already enabled"))
+        }
+
+        if (user.passwordLastSet == null) {
+            if (user.isSSORequiredUser()) {
+                logger.atInfo().addKeyValue("UserCN", user.cn)
```
hugh-emerson, line 44: Really minor, and maybe not an issue, but something like UserCN might be interpreted as the requesting user rather than the target user. 

AdminEmailController adds a key of "userCNToSendEmailTo" to some logs. I think maybe targetUserCN or similar might be fine, but 🤷‍♂️

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEnableDisableUserController.kt
```diff
@@ -0,0 +1,101 @@
...
+    private val userLookupService: UserLookupService,
+    private val userService: UserService,
+    private val setPasswordTokenService: SetPasswordTokenService,
+    private val auditService: UserAuditService,
+) : AdminUserController(userLookupService) {
+
+    private val logger = LoggerFactory.getLogger(javaClass)
+
+    private val allowedRoles =
+        arrayOf(DeltaConfig.DATAMART_DELTA_READ_ONLY_ADMIN, DeltaConfig.DATAMART_DELTA_ADMIN)
```
hugh-emerson, line 31: "Read only" admin shouldn't be allowed to this should they?

hugh-emerson (commented):
The bulk user action for "Access control" lets you specify enabled/disabled notifications. Do we not need to support that?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEnableDisableUserController.kt
BenRamchandani: Makes sense, I'll change it

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEnableDisableUserController.kt
BenRamchandani: Good question, in practice "Read only admin" is assigned to the helpdesk (Excela). They definitely can and do send activation emails, which lets users enable their own accounts, so I think they may as well be able to directly enable and disable accounts. We definitely need to change the name of the role though.

BenRamchandani:
> The bulk user action for "Access control" lets you specify enabled/disabled notifications. Do we not need to support that?
> 
> I guess that's covered by DT-896?

Yeah, for now that's still done in Delta. Definitely want to move that over too, but I've left that for DT-896 (which I've now updated)

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/AdminEditUserControllerTest.kt
BenRamchandani: https://github.com/communitiesuk/delta-auth-service/pull/102

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEnableDisableUserController.kt
hugh-emerson: Cool. Maybe a comment against the enum for now or something.

hugh-emerson (approved):

PR #102, opened 2024-03-11
Refactor bearer token routing

BenRamchandani:


auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/AdminEnableDisableUserControllerTest.kt
```diff
@@ -243,16 +244,14 @@ class AdminEnableDisableUserControllerTest {
...
-                            mockk(relaxed = true),
-                            mockk(relaxed = true),
-                            mockk(relaxed = true),
-                            mockk(relaxed = true),
-                            mockk(relaxed = true),
-                            mockk(relaxed = true),
-                            controller,
-                        )
+                        withBearerTokenAuth {
+                            post("/bearer/admin/enable-user") {
```
PhoebeJackson1, line 248: I imagine it makes no difference but since we are refactoring for consistency anyway I think in some places you have removed bearer from this bit and the urls in the test but in others you haven't? I think without looks tidier personally?

PhoebeJackson1 (approved):
This is so much nicer!

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/AdminEnableDisableUserControllerTest.kt
BenRamchandani: Done, I agree it looks neater :)

PR #109, opened 2024-03-21
DT-694 edit organisations endpoint and tests

JB-softwire:
Edit own organisations endpoint

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
@@ -33,7 +33,7 @@ class EditOrganisationsController(
         val callingUser = userLookupService.lookupUserByCn(session.userCn)
         logger.atInfo().log("Updating organisations for user {}", session.userCn)
 
-        val requestedOrganisations = call.receive<DeltaUserOrganisations>().userSelectedOrgs
+        val requestedOrganisations = call.receive<DeltaUserOrganisations>().selectedDomainOrganisationCodes
         val userDomainOrgs =
             if (callingUser.email == null) setOf() else organisationService.findAllByDomain(callingUser.email)
                 .associateBy { it.code }.keys
@@ -101,6 +101,6 @@ class EditOrganisationsController(
 
     @Serializable
     data class DeltaUserOrganisations(
-        val userSelectedOrgs: List<String>,
+        val selectedDomainOrganisationCodes: List<String>,
     )
 }
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
@@ -1,0 +1,108 @@
+package uk.gov.communities.delta.auth.controllers.internal
+
+import io.ktor.http.*
+import io.ktor.server.application.*
+import io.ktor.server.auth.*
+import io.ktor.server.request.*
+import io.ktor.server.response.*
+import io.ktor.server.routing.*
+import kotlinx.serialization.Serializable
+import org.slf4j.LoggerFactory
+import uk.gov.communities.delta.auth.config.DeltaConfig
+import uk.gov.communities.delta.auth.plugins.ApiError
+import uk.gov.communities.delta.auth.repositories.LdapUser
+import uk.gov.communities.delta.auth.services.*
+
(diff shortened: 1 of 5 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
```diff
@@ -0,0 +1,108 @@
...
+        val callingUser = userLookupService.lookupUserByCn(session.userCn)
+        logger.atInfo().log("Updating organisations for user {}", session.userCn)
+
+        val requestedOrganisations = call.receive<DeltaUserOrganisations>().userSelectedOrgs
+        val userDomainOrgs =
+            if (callingUser.email == null) mapOf() else organisationService.findAllByDomain(callingUser.email)
+                .associateBy { it.code }
+        validateOrganisationRequest(requestedOrganisations, userDomainOrgs)
+
+        val existingOrganisations = getExistingOrganisationsForUser(callingUser)
```
BenRamchandani, line 43: If we're using memberOfToDeltaRolesMapper anyway we may as well get the current organisations from there

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
```diff
@@ -0,0 +1,108 @@
...
+
+        val orgsToAdd = requestedOrganisations.toSet().minus(existingOrganisations.toSet())
+        val orgsToRemove = existingOrganisations.toSet().minus(requestedOrganisations.toSet())
```
BenRamchandani, line 46: We shouldn't remove the user from non-domain organisations, admins can assign users extra organisations that don't match there domain.

We could:
* Just ignore non-domain orgs here
* Have separate organisationsToAdd and organisationsToRemove passed in
* Have separate endpoints that add or remove one organisation

To be honest I think the last option might be nicest? Given the number of organisations is normally very small. Up to you though, if we're just ignoring them we should make it very clear that's what this endpoint does.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
```diff
@@ -0,0 +1,108 @@
...
+        for (org in orgsToAdd) {
+            for (role in userRoles) {
+                val roleGroupString = role.role.adCn(org)
+                groupService.addUserToGroup(callingUser.cn, callingUser.dn, roleGroupString, call, null)
+            }
+        }
+
+        for (org in orgsToRemove) {
+            for (group in callingUser.memberOfCNs) {
+                if (group.endsWith(org)) {
```
BenRamchandani, line 61:
```suggestion
                if (group.endsWith("-$org")) {
```

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/EditOrganisationsControllerTest.kt
```diff
@@ -192,6 +162,42 @@ class EditOrganisationsControllerTest {
         }
     }
 
+    @Test
+    fun userCannotRemoveAllOrganisations() = testSuspend {
+        Assert.assertThrows(ApiError::class.java) {
+            runBlocking {
```
BenRamchandani, line 168: I don't think these need `runBlocking`/`testSuspend`

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
```diff
@@ -27,27 +25,32 @@ class EditOrganisationsController(
...
-    // the user should be a member after the execution of the endpoint
+    // This endpoint takes a list of organisation codes which should contain all and only the domain organisations of which
+    // the user should be a member after the execution of the endpoint. Note that if the user is a member of any organisations
+    // not in their domain, these will NOT be affected by this endpoint.
     private suspend fun updateUserOrganisations(call: ApplicationCall) {
         val session = call.principal<OAuthSession>()!!
         val callingUser = userLookupService.lookupUserByCn(session.userCn)
         logger.atInfo().log("Updating organisations for user {}", session.userCn)
 
         val requestedOrganisations = call.receive<DeltaUserOrganisations>().userSelectedOrgs
```
BenRamchandani, line 36: Maybe worth renaming `userSelectedOrgs` in the request body too, `selectedDomainOrganisationCodes`?

BenRamchandani (approved):

PR #112, opened 2024-03-26
Dt 694

JB-softwire:
Edit username and details endpoints.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditEmailController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditEmailController.kt
@@ -2,42 +2,45 @@ package uk.gov.communities.delta.auth.controllers.internal
 
 import io.ktor.http.*
 import io.ktor.server.application.*
-import io.ktor.server.auth.*
 import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
 import kotlinx.serialization.Serializable
 import org.slf4j.LoggerFactory
 import uk.gov.communities.delta.auth.plugins.ApiError
-import uk.gov.communities.delta.auth.services.OAuthSession
+import uk.gov.communities.delta.auth.services.DeltaSystemRole
 import uk.gov.communities.delta.auth.services.UserLookupService
 import uk.gov.communities.delta.auth.services.UserService
 
-class EditUsernameController(
+class AdminEditEmailController(
     private val userLookupService: UserLookupService,
     private val userService: UserService,
-) {
+) : AdminUserController(userLookupService) {
 
     private val logger = LoggerFactory.getLogger(javaClass)
 
     fun route(route: Route) {
-        route.post { updateUsername(call) }
+        route.post { updateEmail(call) }
     }
 
-    private suspend fun updateUsername(call: ApplicationCall) {
-        val session = call.principal<OAuthSession>()!!
-        val callingUser = userLookupService.lookupUserByCn(session.userCn)
-        logger.atInfo().log("Updating organisations for user {}", session.userCn)
+    private suspend fun updateEmail(call: ApplicationCall) {
+        val session = getSessionIfUserHasPermittedRole(
+            arrayOf(DeltaSystemRole.ADMIN), call
+        )
+        val requestData = call.receive<DeltaEmailChangeRequest>()
+        val requestedEmail = requestData.newEmail
 
-        val requestedUsername = call.receive<DeltaUsername>().username
+        val userToEdit = userLookupService.lookupUserByCn(requestData.userToEditCn)
 
-        validateUsername(requestedUsername)
+        logger.atInfo().log("Updating email and username for user {}", userToEdit.cn)
 
-        userService.updateUsername(callingUser, requestedUsername, null, call)
+        validateEmail(requestedEmail)
+
+        userService.updateEmail(userToEdit, requestedEmail, session, call)
         return call.respond(mapOf("message" to "Username has been updated."))
     }
 
-    private fun validateUsername(requestedUsername: String) {
+    private fun validateEmail(requestedUsername: String) {
         if (requestedUsername.isBlank()) {
             throw ApiError(
                 HttpStatusCode.Forbidden,
@@ -48,7 +51,8 @@ class EditUsernameController(
     }
 
     @Serializable
-    data class DeltaUsername(
-        val username: String,
+    data class DeltaEmailChangeRequest(
+        val userToEditCn: String,
+        val newEmail: String,
     )
 }
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditEmailController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditEmailController.kt
@@ -5,6 +5,7 @@ import io.ktor.server.application.*
 import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
+import kotlinx.serialization.SerialName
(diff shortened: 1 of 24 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
```diff
@@ -59,12 +123,67 @@ class EditAccessGroupsController(
...
+        groupName: String,
+        callingUser: LdapUser,
+        targetUser: LdapUser,
+        allAccessGroupNames: List<String>,
+    ) {
+        validateGroupExistenceAndUserAuthority(allAccessGroupNames, groupName, callingUser)
+
+        if (targetUser.memberOfCNs.contains(LDAPConfig.DATAMART_DELTA_PREFIX + groupName)) {
+            throw ApiError(
+                HttpStatusCode.Forbidden,
```
BenRamchandani, line 141: Nit most of these should be 400 (or maybe 409 or 422), providing different credentials is never going to make this request succeed.

BenRamchandani (commented):
This looks good, I'll have another proper look next week

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditEmailController.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.ktor.server.routing.*
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import org.slf4j.LoggerFactory
+import uk.gov.communities.delta.auth.plugins.ApiError
+import uk.gov.communities.delta.auth.services.DeltaSystemRole
+import uk.gov.communities.delta.auth.services.UserLookupService
+import uk.gov.communities.delta.auth.services.UserService
+
+class AdminEditEmailController(
```
BenRamchandani, line 16: Maybe `AdminEditUserEmailController`? Just to make it clearer editing a user, not an email

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditEmailController.kt
```diff
@@ -0,0 +1,59 @@
...
+            arrayOf(DeltaSystemRole.ADMIN), call
+        )
+        val requestData = call.receive<DeltaEmailChangeRequest>()
+        val requestedEmail = requestData.newEmail
+
+        val userToEdit = userLookupService.lookupUserByCn(requestData.userToEditCn)
+
+        logger.atInfo().log("Updating email and username for user {}", userToEdit.cn)
+
+        validateEmail(requestedEmail)
```
BenRamchandani, line 38: Let's log lots of detail

```suggestion
        validateEmail(requestedEmail)
        logger.atInfo().addKeyValue("oldUserEmail", userToEdit.email)
            .addKeyValue("oldUserCN", userToEdit.cn)
            .addKeyValue("newUserEmail", requestedEmail)
            .log("Request to update email and username for user")
```

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditEmailController.kt
```diff
@@ -0,0 +1,59 @@
...
+
+        userService.updateEmail(userToEdit, requestedEmail, session, call)
+        return call.respond(mapOf("message" to "Username has been updated."))
+    }
+
+    private fun validateEmail(requestedUsername: String) {
```
BenRamchandani, line 44: Maybe we should check it's a valid email as well. I know we don't strictly require it, but in practice for editing users in the frontend I think it's fine to. There's an `EmailAddressChecker` you can use

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
```diff
@@ -59,12 +123,67 @@ class EditAccessGroupsController(
...
         val accessGroupActions = generateAccessGroupActions(accessGroupRequestMap, currentAccessGroups, selectedOrgs)
 
-        executeAccessGroupActions(accessGroupActions, session, callingUser, call)
+        executeAccessGroupActions(accessGroupActions, null, callingUser, call)
 
         return call.respond(mapOf("message" to "Access groups have been updated. Any changes to your roles or access groups will take effect the next time you log in."))
     }
 
-    private suspend fun validateRequest(
+    fun validateAddAccessGroupRequest(
```
BenRamchandani, line 131: I imagine the grant wizard also adds the user to `dclg` in the access group, can you check?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
```diff
@@ -28,22 +29,85 @@ class EditAccessGroupsController(
...
+
+        val allAccessGroups = accessGroupsService.getAllAccessGroups()
+        validateAddAccessGroupRequest(targetGroupName, callingUser, targetUser, allAccessGroups.map { it.name })
+
+        groupService.addUserToGroup(
+            targetUser.cn,
+            targetUser.dn,
+            targetGroupADName,
+            call,
+            null,
```
BenRamchandani, line 55: We should pass in the session here and below

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
JB-softwire: We determined that it does but can you explain what you had in mind with this? After discussion with Hugh I removed the requirement that users not already be in/out of the group they are being added to/removed from.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditUserEmailController.kt
```diff
@@ -0,0 +1,73 @@
...
+    private suspend fun updateEmail(call: ApplicationCall) {
+        val session = getSessionIfUserHasPermittedRole(
+            arrayOf(DeltaSystemRole.ADMIN), call
+        )
+        val requestData = call.receive<DeltaEmailChangeRequest>()
+        val requestedEmail = requestData.newEmail
+
+        val userToEdit = userLookupService.lookupUserByCn(requestData.userToEditCn)
+
+        logger.atInfo().log("Updating email and username for user {}", userToEdit.cn)
```
BenRamchandani, line 38: Don't need this too

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
BenRamchandani: Nothing specific, but we'll need to make sure we keep that behaviour

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/EditUserAccessGroupsControllerTest.kt
```diff
@@ -235,13 +237,122 @@ class EditUserAccessGroupsControllerTest {
...
+            }
+            coVerify(exactly = 0) {
+                groupService.removeUserFromGroup(
+                    any(),
+                    any(),
+                    any(),
+                    any(),
+                    any()
+                )
+            }
```
BenRamchandani, line 280: Fine to keep it for consistency with other tests if you like, but the `confirmVerified` makes it unnecessary

BenRamchandani (approved):

PR #118, opened 2024-04-03
Add memberOf to role mapping to userLookupService

BenRamchandani:
Add a new function to UserLookupService that reads the user from AD and then interprets its roles, access groups etc.

This simplifies a few code paths that previously had to read all organisations and access groups and then call the roles mapper. It hasn't helped as much as I'd hoped, but it should also be easier to migrate to a database where we wouldn't need the list of all organisations and access groups to load a user's roles.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminGetUserController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminGetUserController.kt
@@ -4,11 +4,9 @@ import io.ktor.http.*
 import io.ktor.server.application.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
-import kotlinx.serialization.Serializable
 import org.slf4j.LoggerFactory
-import uk.gov.communities.delta.auth.repositories.LdapUser
 import uk.gov.communities.delta.auth.services.DeltaSystemRole
-import uk.gov.communities.delta.auth.services.MemberOfToDeltaRolesMapper
+import uk.gov.communities.delta.auth.services.LdapUserWithRoles
 import uk.gov.communities.delta.auth.services.UserLookupService
 import javax.naming.NameNotFoundException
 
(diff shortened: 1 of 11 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminGetUserController.kt
```diff
@@ -32,22 +29,14 @@ class AdminGetUserController(
...
-            val allOrganisations = async { organisationService.findAllNamesAndCodes() }
-            val allAccessGroups = async { accessGroupsService.getAllAccessGroups() }
-
-            val roles = memberOfToDeltaRolesMapperFactory(
-                user.cn, allOrganisations.await(), allAccessGroups.await()
-            ).map(user.memberOfCNs)
-            call.respond(UserWithRoles(user, roles))
-        }
-
+        call.respond(UserWithRoles(user.user, user.roles))
```
hugh-emerson, line 39: Isn't this just `call.respond(user)`?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
```diff
@@ -30,41 +31,39 @@ class EditOrganisationsController(
     // not in their domain, these will NOT be affected by this endpoint.
     private suspend fun updateUserOrganisations(call: ApplicationCall) {
         val session = call.principal<OAuthSession>()!!
-        val callingUser = userLookupService.lookupUserByCn(session.userCn)
+        val callingUser = userLookupService.lookupUserByCNAndLoadRoles(session.userCn)
```
hugh-emerson, line 34: I think it's *slightly* unintuitive that variables called "user" or "callingUser" are now UserWithRoles objects instead of LdapUser objects. So to actually get the user, you have to call user.user

Here, would it be nicer to do something like `val (callingUser, callingUserRoles) = userLookupService.lookupUserByCNAndLoadRoles(session.userCn)`?

hugh-emerson (approved):
Seems like a sensible refactor.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminGetUserController.kt
BenRamchandani: They're different classes with the same name, identical except this on is `@Serializable`, I'll commonise them

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditOrganisationsController.kt
BenRamchandani: Yeah that looks much nicer :+1:

PR #123, opened 2024-04-15
DT-696 Send notification emails when user added to dclg access group

BenRamchandani:
**Description** Use the auth service to send notification emails when a user is added to the dclg organisation in an access group.
There's no nice abstraction layer to have a single point to put this (and I couldn't find a nice way to add one, the admin edit user endpoint wants to work at the AD group level really so it can fix up existing bad group membership), so it's something else the controllers have to do.

**Local testing** I've done some manual testing and made the existing unit tests pass, but there's more automated testing to do.

**Release** We'll need to set the feature flag in Terraform to enable this, and confirm a list of recipients for staging


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/AccessGroupRole.kt
```diff
@@ -0,0 +1,12 @@
+package uk.gov.communities.delta.auth.services
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class AccessGroupRole(
```
BenRamchandani, line 6: Moved out from inside MemberOfToDeltaRolesMapper

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/AdminUserCreationControllerTest.kt
```diff
@@ -299,6 +317,14 @@ class AdminUserCreationControllerTest {
             AccessGroup("access-group-1", "STATS", "access group 1", false, false),
             AccessGroup("access-group-2", "STATS", "access group 2", false, false),
         )
+        every {
+            accessGroupDCLGMembershipUpdateEmailService.sendNotificationEmailsForUserChange(
+                any(),
```
PhoebeJackson1, line 322: (Can these be on one line?)

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/AccessGroupDCLGMembershipUpdateEmailService.kt
```diff
@@ -0,0 +1,160 @@
...
+    private val emailConfig: EmailConfig,
+) {
+    private val supervisorJob = SupervisorJob()
+    private val logger = LoggerFactory.getLogger(javaClass)
+    private val emailChecker = EmailAddressChecker()
+
+    private val defaultRecipients = emailConfig.dclgAccessGroupUpdateAdditionalRecipients.map {
+        EmailRecipient(
+            it,
+            it.substringBefore('@')
```
PhoebeJackson1, line 38: I feel like I've seen emails get caught in our spam filter for being addressed to e.g. "phoebe.jackson" ? Not entirely sure and can't find any proof but might be worth keeping an eye out for?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/AccessGroupDCLGMembershipUpdateEmailService.kt
```diff
@@ -0,0 +1,160 @@
...
+        user: UpdatedUser,
+        actingUser: LdapUser,
+        previousAccessGroups: Collection<AccessGroupRole>,
+        newAccessGroups: Collection<AccessGroupRole>,
+    ) {
+        if (!emailConfig.dclgAccessGroupUpdateNotificationsEnabled) {
+            return
+        }
+
+        val previousDclgGroups = previousAccessGroups.filter { it.organisationIds.contains("dclg") }.toSet()
```
PhoebeJackson1, line 75: Bit of inconsistency between capitalising "dclg" in stuff (I think caps is easier to read personally 🤷 )

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/EditAccessGroupsControllerTest.kt
```diff
@@ -402,19 +516,26 @@ class EditAccessGroupsControllerTest {
...
         coEvery { organisationService.findAllByEmail(externalUser.email) } returns listOf(
             Organisation("orgCode1", "Organisation Name 1"),
             Organisation("orgCode2", "Organisation Name 2"),
             Organisation("orgCode3", "Organisation Name 3"),
             Organisation("orgCode4", "Organisation Name 4"),
         )
+        coEvery { organisationService.findAllByEmail(internalUser.email) } returns listOf(
+            Organisation("orgCode1", "Organisation Name 1"),
+            Organisation("orgCode2", "Organisation Name 2"),
+            Organisation("dclg", "The Department"),
```
PhoebeJackson1, line 537: 😂

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/AccessGroupsService.kt
```diff
@@ -18,9 +18,9 @@ data class AccessGroup(
     val registrationDisplayName: String?,
     val enableOnlineRegistration: Boolean,
     val enableInternalUser: Boolean,
-    val isDelegate: Boolean? = null,
```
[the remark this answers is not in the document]
BenRamchandani: It's just unused here, I think previously this "AccessGroup" class was sometimes used represent a user's access group, but now "AccessGroupRole" is always used for that (which does have an `isDelegate`) and this class represents the properties of an access group loaded from AD independent of any user.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/EmailService.kt
```diff
@@ -236,20 +236,54 @@ class EmailService(
 
         logger.atInfo().addKeyValue("userCN", userCN).log("Sent reset-password email")
     }
+
+    suspend fun sendDLUHCUserAddedToUserGroupEmail(
```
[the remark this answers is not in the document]
BenRamchandani: I think that makes sense, I've updated the template and function names to hopefully be clearer, and added a comment.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/AccessGroupDCLGMembershipUpdateEmailService.kt
```diff
@@ -0,0 +1,160 @@
...
+                    "Failed to send notification emails for change to user {}, they were added to groups {}",
+                    user.email,
+                    addedGroups.map { it },
+                    e,
+                )
+            }
+        }
+    }
+
+    private suspend fun sendEmailsForChangeToAccessGroup(
```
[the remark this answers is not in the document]
BenRamchandani: Adding only, renamed

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/AccessGroupDCLGMembershipUpdateEmailService.kt
BenRamchandani: Ah good to know. This will only be used for the "additional recipients" like DeltaAdmin@levellingup.gov.uk so will hopefully be OK.

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/EditAccessGroupsControllerTest.kt
BenRamchandani: It's too long to type out 😂

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/DeltaUserDetailsRequest.kt
```diff
@@ -65,12 +63,16 @@ data class DeltaUserPermissions(
                 ?.let { throw BadInput("Cannot be delegate of access group $it without being a member") }
 
             val accessGroups = userDetailsRequest.accessGroups.map { prefixedAccessGroupName ->
+                val accessGroupName = prefixedAccessGroupName.removePrefix(LDAPConfig.DATAMART_DELTA_PREFIX)
+                val accessGroup = allAccessGroups[accessGroupName]!!
```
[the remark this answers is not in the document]
BenRamchandani: In terms of the `!!`, yes it's checked above. Whether we should require it, I think you're right not everywhere does at the moment, but I think it probably makes sense to. I feel like a new access group in a bulk user upload is more likely to be a mistake than intentional

auth-service/src/main/resources/templates/thymeleaf/emails/dluhc-user-added-to-collection.html
```diff
@@ -0,0 +1,46 @@
...
+<!--/*@thymesVar id="userEmailAddress" type="java.lang.String" */-->
+<!--/*@thymesVar id="accessGroupName" type="java.lang.String" */-->
+<!--/*@thymesVar id="actingUserEmail" type="java.lang.String" */-->
+
+<p>
+    This is to notify you that <span th:text="${userFullName}">#userFullName#</span>
+    (<span th:text="${userEmailAddress}">user@example.com</span>)
+    has
+    <span
+        th:text="${actingUserEmail == userEmailAddress} ? 'added themselves to this collection' : 'been added to this collection by ' + ${actingUserEmail}"></span>.
```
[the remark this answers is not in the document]
BenRamchandani: Updated the text to include it

PhoebeJackson1 (approved):

PR #129, opened 2024-04-22
DT-696 Terraform configuration for notification emails

BenRamchandani:
**Description** Add terraform config for the DCLG access group notifications feature. Currently enabled on test and staging, sending emails only to us for testing for now.

**Local testing** Applied to test

**Release** No extra steps


terraform/staging/main.tf
```diff
@@ -31,6 +31,10 @@ data "terraform_remote_state" "common_infra" {
 locals {
   environment                    = "staging"
   cloudwatch_log_expiration_days = 30
+  dclg_access_group_notification_settings = {
+    enabled                     = true
+    additional_recipient_emails = ["Group-DLUHCDeltaDevNotifications+staging@softwire.com"]
```
hugh-emerson, line 36: Should start including `delta-notifications@levellingup.gov.uk` for stuff like this

auth-service/src/main/resources/templates/thymeleaf/emails/user-added-to-dclg-in-access-group.html
```diff
@@ -24,16 +24,11 @@
...
-    Please contact <a th:href="'mailto:' + ${userEmailAddress}"><span
-    th:text="${userFullName}">#userFullName#</span></a>
-    or <a href="mailto:DELTAAdmin@levellingup.gov.uk">DELTAAdmin@levellingup.gov.uk</a> if you think that they should
-    not have access to this collection.
 </p>
 
 <p>
     If you have any questions or need help please email the DLUHC Service Desk on <a
-    href="mailto:dluhc.digital-services@levellingup.gov.uk">dluhc.digital-services@levellingup.gov.uk</a> or call
-    0203 829 0743.
```
hugh-emerson, line 36: I'd leave the number 🤷‍♂️ 

It appears in a lot of places, so if we want to get rid of it, it's a slightly bigger change. I assume it goes to Exela.

hugh-emerson (changes requested):

hugh-emerson (approved):

PR #130, opened 2024-04-23
DT-983 new endpoint for MFA reset plus tests

JB-softwire:
**Description** Auth service side of putting MFA reset in the auth service.

**Local testing** Local testing for controller, and service method worked in E2E testing.

**Release** No extra release steps or documentation changes that I am aware of.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
@@ -300,6 +300,23 @@ class UserService(
         }
     }
 
+    suspend fun resetMfaToken(userToReset: LdapUser, session: OAuthSession, call: ApplicationCall) {
+        // the TOTP secret is stored in the "unixHomeDirectory" Active Directory field
+        try {
+            val modificationItem = ModificationItem(DirContext.REMOVE_ATTRIBUTE, BasicAttribute("unixHomeDirectory", null))
+            val modificationArray = arrayOf(modificationItem)
+            ldapServiceUserBind.useServiceUserBind {
+                it.modifyAttributes(userToReset.dn, modificationArray)
+                logger.atInfo().addKeyValue("unixHomeDirectory", "").log("MFA token deleted")
+            }
+        } catch (e: Exception) {
+            logger.atError().addKeyValue("unixHomeDirectory", "").log("Error deleted MFA token", e)
+            throw e
+        }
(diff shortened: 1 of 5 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
```diff
@@ -300,6 +300,23 @@ class UserService(
...
+        // the TOTP secret is stored in the "unixHomeDirectory" Active Directory field
+        try {
+            val modificationItem = ModificationItem(DirContext.REMOVE_ATTRIBUTE, BasicAttribute("unixHomeDirectory", null))
+            val modificationArray = arrayOf(modificationItem)
+            ldapServiceUserBind.useServiceUserBind {
+                it.modifyAttributes(userToReset.dn, modificationArray)
+                logger.atInfo().addKeyValue("unixHomeDirectory", "").log("MFA token deleted")
+            }
+        } catch (e: Exception) {
+            logger.atError().addKeyValue("unixHomeDirectory", "").log("Error deleted MFA token", e)
```
BenRamchandani, line 313: Again minor, but I'd either get rid of the try catch, or at least log the username of the user we're updating. If you were coming to investigate this error you'd already know we're setting "unixHomeDirectory" to be empty from the stack trace (we also know the target user, but only because the controller logs it, it wouldn't be part of this message)

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/services/UserService.kt
```diff
@@ -300,6 +300,23 @@ class UserService(
...
     }
 
+    suspend fun resetMfaToken(userToReset: LdapUser, session: OAuthSession, call: ApplicationCall) {
+        // the TOTP secret is stored in the "unixHomeDirectory" Active Directory field
+        try {
+            val modificationItem = ModificationItem(DirContext.REMOVE_ATTRIBUTE, BasicAttribute("unixHomeDirectory", null))
+            val modificationArray = arrayOf(modificationItem)
+            ldapServiceUserBind.useServiceUserBind {
+                it.modifyAttributes(userToReset.dn, modificationArray)
+                logger.atInfo().addKeyValue("unixHomeDirectory", "").log("MFA token deleted")
```
BenRamchandani, line 310:
```suggestion
                logger.atInfo().addKeyValue("userToResetCN", userToReset.cn).log("MFA token deleted")
```

BenRamchandani (approved):

PR #137, opened 2024-05-17
DT-629 Enable user update notification emails on production

BenRamchandani:
**Description** Enable the new user update notifications on production. Should be roughly co-ordinated with removing the old `mhclg-user-added-to-stats-user-group` part of the ldap-audit which I'll raise a ticket for, though I think it's fine if both are enabled for a bit.

### DIFF
--- a/terraform/production/main.tf
+++ b/terraform/production/main.tf
@@ -32,8 +32,8 @@ locals {
   environment                    = "production"
   cloudwatch_log_expiration_days = 731
   dclg_access_group_notification_settings = {
-    enabled                     = false
-    additional_recipient_emails = []
+    enabled                     = true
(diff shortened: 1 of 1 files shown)


terraform/production/main.tf
```diff
@@ -32,8 +32,8 @@ locals {
   environment                    = "production"
   cloudwatch_log_expiration_days = 731
   dclg_access_group_notification_settings = {
-    enabled                     = false
-    additional_recipient_emails = []
+    enabled                     = true
+    additional_recipient_emails = ["Group-DLUHCDeltaNotifications@softwire.com", "DELTAadmin@levellingup.gov.uk", "DELTAStatSupport@levellingup.gov.uk"]
```
hugh-emerson, line 36: If we're including Softwire's notifications address, we should include delta-notifications@levellingup.gov.uk as well.

hugh-emerson (approved):

PR #141, opened 2024-05-20
DT-1013: add endpoint for managing LDAP groups, to replace internal app

hugh-emerson:
**Description** Add an endpoint for adding users to LDAP groups.
The ticket suggested taking an organisation and an access group, but I don't see the point given that the ML layer already creates the group CN?

I guess yeah, ideally other systems wouldn't know about the implementation details but not convinced it's worth worrying about at this stage.

**Local testing** Still doing some testing locally

**Release** Must be released before Delta side


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
```diff
@@ -328,7 +329,29 @@ fun Route.bearerTokenRoutes(
...
 }
 
-fun Route.serviceUserRoutes(samlTokenController: GenerateSAMLTokenController) {
+fun Route.serviceUserRoutes(
+    samlTokenController: GenerateSAMLTokenController,
+    editLdapGroupsController: EditLdapGroupsController
+) {
+    samlTokenRoutes(samlTokenController)
+
+    authenticate(DELTA_AD_LDAP_SERVICE_USERS_AUTH_NAME, strategy = AuthenticationStrategy.Required) {
```
BenRamchandani, line 338: Can use the client header auth too (i.e. authenticating which service is making the call) if we send it from MarkLogic
```
map:entry("Delta-Client", $config:SAML-TOKEN-GENERATOR-CLIENT-ID || ":" || $config:SAML-TOKEN-GENERATOR-CLIENT-SECRET),
```
in the header map like saml-token.xqy. Doesn't really matter here since we're only expecting MarkLogic to call it, up to you.

hugh-emerson: I think just basic auth is more intuitive (and less work)

BenRamchandani (approved):

PR #142, opened 2024-05-21
DT-976: Use userGUID instead of userCN (Part 1)

PhoebeJackson1:
https://dluhcdigital.atlassian.net/browse/DT-976

**Description** Code for the first release for DT-976 which uses the userGUID instead of userCN for all possible auth service uses.

- Adds relevant guid columns to all database tables
- Stores guid in all places where rows are added to those tables
- Adds in checks for userGUID being passed as a parameter in requests (set up ground work and TODOs (labelled DT-1022) for [DT-1022](https://dluhcdigital.atlassian.net/browse/DT-1022)
- Added TODOs (labelled DT-976-2) for the second part of this ticket to be released once the database table guid values have been backfilled
- Made create user return the created user and used it where possible/relevant
- NB: There is a lot of passing around of the userLookupService added in this ticket so that the GUID can be found in cases were it currently isn't available (particularly from the session), this won't be used very much but allows for backwards compatability and thus no downtime and will be removed in the second PR of this ticket
- Allow user look up from session directly (based on GUID once/if it exists)
- Translates userGUID from java UUID back to active directory format
- Adds lookup by guid in AD
- All tests passing

**Local testing** I have locally tested stuff as I went along including error handling and the process for no downtime e.g. sessions that do not have guid in the table still work and can do everything

Testing directions:
Test all flows that use the auth service checking that they behave as expected and are auditted correctly with GUIDs:
1. Login flow
2. SSO login flow
3. Forgot password
4. User creation (including setting password) as self
5. User creation as admin
6. User creation via SSO
7. Reset password from admin call
8. Edit user
9. Enable/disable users
10. Reset MFA
11. Enable/disable notifications
12. Change own access groups
13. Change username (NB: this will confuse the audit still until 2nd release)

**Release** 
Auth service release only, will be followed by updating userGUID map and then second release
No changes needed for local set up until second release
No special actions needed
Updated documentation: it will mean that some places log guid rather than CN, currently we don't have a way to find which user a guid belongs to easily, raised https://dluhcdigital.atlassian.net/browse/DT-1043 to solve this - should we have a workaround for the meanwhile? It will likely involve translating the guid which is pretty messy?
Second release tracked by https://dluhcdigital.atlassian.net/browse/DT-1044

[DT-1022]: https://dluhcdigital.atlassian.net/browse/DT-1022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/FetchUserAuditController.kt
```diff
@@ -34,14 +34,19 @@ class FetchUserAuditController(
     private suspend fun ApplicationCall.getUserAuditJson() {
         val session = principal<OAuthSession>()!!
         val page = parameters["page"]?.toInt()
+        // TODO DT-1022 - send GUID instead of CN
```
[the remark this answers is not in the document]
BenRamchandani: Yeah we could remove the actual user CN from this JSON response. I think we don't want to be doing potentially lots of AD lookups to get all the editing users, it would be alright for here, but not for the CSV where there could be a lot of unique users.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/config/LDAPConfig.kt
```diff
@@ -41,7 +41,8 @@ data class LDAPConfig(
         )
 
         val VALID_EMAIL_REGEX = Regex("^[\\w\\-+.']+@([\\w\\-']+\\.)+[\\w\\-]{2,4}$")
-        val VALID_USERNAME_REGEX = Regex("^[\\w\\-+.!']+$")
+        val VALID_USER_CN_REGEX = Regex("^[\\w\\-+.!']+$")
+        val VALID_USER_GUID_REGEX = Regex("^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$")
```
PhoebeJackson1, line 45: Note to self: Change this to 
`"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"`

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/AdminEditUserController.kt
```diff
@@ -35,22 +35,18 @@ class AdminEditUserController(
...
+            "Something went wrong, please try again",
+            "update_user_as_admin"
+        )
+        val userToUpdateRoles = userLookupService.loadUserRoles(userToUpdate).roles // TODO DT-1022 - use lookupUserByGUIDAndLoadRoles
 
+        logger.atInfo().log("Updating user with GUID ${userToUpdate.getGUID()}")
         val updatedDeltaUserDetailsRequest = call.receive<DeltaUserDetailsRequest>()
 
-        if (updatedDeltaUserDetailsRequest.id.replace("@", "!") != cn) throw ApiError(
+        if (updatedDeltaUserDetailsRequest.id.replace("@", "!") != userToUpdate.cn) throw ApiError(
```
keithkennedy, line 49: Should this use `LDAPConfig.emailToCN`?

PhoebeJackson1: Yeah, that wasn't something I changed in this PR but I've now checked for other uses and changed it in 4 places total :)

BenRamchandani (approved):

PR #144, opened 2024-05-29
DT-984: May patching

YD-DLUHC:
**Description** 

May patching

### DIFF
--- a/auth-service/build.gradle.kts
+++ b/auth-service/build.gradle.kts
@@ -60,19 +60,19 @@ dependencies {
 
     // Metrics
     implementation("io.ktor:ktor-server-metrics-micrometer:$ktorVersion")
-    implementation("io.micrometer:micrometer-registry-cloudwatch2:1.12.4")
+    implementation("io.micrometer:micrometer-registry-cloudwatch2:1.13.0")
 
     //Emails
     implementation("com.sun.mail:jakarta.mail:2.0.1")
 
     // Logging
-    implementation("ch.qos.logback:logback-classic:1.5.3")
+    implementation("ch.qos.logback:logback-classic:1.5.6")
     implementation("net.logstash.logback:logstash-logback-encoder:7.4") // Structured log encoder
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-slf4j:1.8.0")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-slf4j:1.8.1")
 
     // OpenSAML
-    implementation("org.opensaml:opensaml-core:4.3.0")
-    implementation("org.opensaml:opensaml-saml-impl:4.3.0")
+    implementation("org.opensaml:opensaml-core:4.0.1")
+    implementation("org.opensaml:opensaml-saml-impl:4.0.1")
 
     // Database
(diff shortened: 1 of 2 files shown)


auth-service/build.gradle.kts
```diff
@@ -60,19 +60,19 @@ dependencies {
...
+    implementation("ch.qos.logback:logback-classic:1.5.6")
     implementation("net.logstash.logback:logstash-logback-encoder:7.4") // Structured log encoder
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-slf4j:1.8.0")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-slf4j:1.8.1")
 
     // OpenSAML
-    implementation("org.opensaml:opensaml-core:4.3.0")
-    implementation("org.opensaml:opensaml-saml-impl:4.3.0")
+    implementation("org.opensaml:opensaml-core:4.0.1")
+    implementation("org.opensaml:opensaml-saml-impl:4.0.1")
```
BenRamchandani, line 75: This is a downgrade! 4.3.2 is the latest. https://mvnrepository.com/artifact/org.opensaml/opensaml-core?repo=shibboleth-releases

BenRamchandani (changes requested):
There are new versions of ktor (2.3.11) and Kotlin (1.9.24) too

auth-service/build.gradle.kts
YD-DLUHC: Upgraded versions

auth-service/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
-val ktorVersion = "2.3.9"
-val kotlinVersion = "1.9.23"
+val ktorVersion = "2.3.11"
+val kotlinVersion = "2.0.0"
```
BenRamchandani, line 2: Let's not jump major Kotlin versions yet, can you raise a ticket for it. Make sure the versions just below in the `plugins` block match these

YD-DLUHC: added versions to block below and ticket raised

auth-service/build.gradle.kts
```diff
@@ -60,19 +60,19 @@ dependencies {
...
+    implementation("ch.qos.logback:logback-classic:1.5.6")
     implementation("net.logstash.logback:logstash-logback-encoder:7.4") // Structured log encoder
-    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-slf4j:1.8.0")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-slf4j:1.8.1")
 
     // OpenSAML
-    implementation("org.opensaml:opensaml-core:4.3.0")
-    implementation("org.opensaml:opensaml-saml-impl:4.3.0")
+    implementation("org.opensaml:opensaml-core:4.3.2")
+    implementation("org.opensaml:opensaml-saml-impl:5.1.2")
```
BenRamchandani, line 75: I think we should leave these matching unless you're reasonably confident 4 and 5 will work together
```suggestion
    implementation("org.opensaml:opensaml-core:4.3.2")
    implementation("org.opensaml:opensaml-saml-impl:4.3.2")
```

BenRamchandani (approved):

PR #145, opened 2024-05-30
DT-976-2: Remove uses of userCN (using userGUID instead)

PhoebeJackson1:
**Description**
(Second half of DT-976, being tracked by DT-1044)
Delete the user cn columns from the database tables
Introduce the user_guid_map table as a way to get user_cn from guid and vice versa without an AD lookup
Uses SQL to get the cns corresponding to the auditted guids so that the audits can still show readable cns but the cns are not auditted -> audit is no longer useless after an email change and also doesn't rely on personal data as the identifying piece of information
Remove the additional code that was introduced in DT-976 (part 1) for the transition period.
Remove any possible uses of userCN, update logs to use userGUID where possible. 
I have raised https://dluhcdigital.atlassian.net/browse/DT-1043 to make the GUID a searchable term on the frontend so that we can easier reverse look up a user from their GUID, until this is done username and guid are both logged by MDC logs.

**Local testing** 
I have tested the audits and register/set password flow as well as lots of other changes along the way. I have run the playwright tests locally and they pass (after a change to the set password url - PR in delta repo to follow)
I have updated all auth service tests and they all pass. I have also added a few tests for untested user flows.

Testing directions:
Test all auth service user flows and ensure they are audited correctly, this should include:

- Login - standard user and SSO
- Edit user - self (non admin), self (admin), other user (as admin)
- Disable/enable user (as admin)
- Reset MFA (as admin)
- New User as admin
- Register self
- SSO registration
- Sending emails as admin
- Forgot password flow
- Updating username (as admin)
- Update user notification status (as admin)
- Impersonating a user (as admin)
- Audits - one user as admin and all users as admin and self (for actions before and after the code was released)

**Release** 

Must be released after (separately from) DT-976 (part 1)

Between the two releases the top 43 lines of the migration (V10) should be run 

UPDATE delta_session
SET user_guid = ugm.newguid
FROM user_guid_map ugm
WHERE delta_session.username = ugm.cn
  AND delta_session.user_guid IS NULL;

UPDATE delta_session
SET impersonated_user_guid = ugm.newguid
FROM user_guid_map ugm
WHERE delta_session.impersonated_user_cn IS NOT NULL
  AND delta_session.impersonated_user_cn = ugm.cn
  AND delta_session.impersonated_user_guid IS NULL;

UPDATE authorization_code
SET user_guid = ugm.newguid
FROM user_guid_map ugm
WHERE authorization_code.username = ugm.cn
  AND authorization_code.user_guid IS NULL;

UPDATE reset_password_tokens
SET user_guid = ugm.newguid
FROM user_guid_map ugm
WHERE reset_password_tokens.user_cn = ugm.cn
  AND reset_password_tokens.user_guid IS NULL;

UPDATE set_password_tokens
SET user_guid = ugm.newguid
FROM user_guid_map ugm
WHERE set_password_tokens.user_cn = ugm.cn
  AND set_password_tokens.user_guid IS NULL;

UPDATE user_audit
SET user_guid = ugm.newguid
FROM user_guid_map ugm
WHERE user_audit.user_cn = ugm.cn
  AND user_audit.user_guid IS NULL;

UPDATE user_audit
SET editing_user_guid = ugm.newguid
FROM user_guid_map ugm
WHERE user_audit.editing_user_cn IS NOT NULL
  AND user_audit.editing_user_guid IS NULL
  AND user_audit.editing_user_cn = ugm.cn;

After running this for the first time the UpdateUserGUIDMap task should be run on production so that it is up to date with the latest versions and then the above SQL should be run again.

Any remaining rows with user_guid (or editing_user_guid etc) that are NULL need to be resolved before the second release. This is because the migration will remove the user cn column from the user audit (and other database tables) there are some user_cns that no longer map to a user_guid since the user's name has been changed. There are ~50 of these on production. So, before this is released (and thus before the full migration is run and the columns deleted) the relevant rows should either be backfilled with the correct user_guid (if we can find it manually) or deleted and noted in the ticket (DT-976) if we can't find it manually.

The SQL to find the list of user_cns for which this is an issue is 

SELECT DISTINCT * FROM (
                           SELECT DISTINCT user_cn
                           FROM user_audit
                                    LEFT JOIN user_guid_map ON user_audit.user_cn = user_guid_map.cn
                           WHERE user_guid_map.cn IS NULL
                           UNION ALL
                           SELECT DISTINCT editing_user_cn
                           FROM user_audit
                                    LEFT JOIN user_guid_map ON user_audit.editing_user_cn = user_guid_map.cn
                           WHERE user_guid_map.cn IS NULL
                           UNION ALL
                           SELECT DISTINCT username
                           FROM delta_session
                                    LEFT JOIN user_guid_map ON delta_session.username = user_guid_map.cn
                           WHERE user_guid_map.cn IS NULL
                           UNION ALL
                           SELECT DISTINCT impersonated_user_cn
                           FROM delta_session
                                    LEFT JOIN user_guid_map ON delta_session.impersonated_user_cn = user_guid_map.cn
                           WHERE user_guid_map.cn IS NULL
                           UNION ALL
                           SELECT DISTINCT username
                           FROM authorization_code
                                    LEFT JOIN user_guid_map ON authorization_code.username = user_guid_map.cn
                           WHERE user_guid_map.cn IS NULL
                           UNION ALL
                           SELECT DISTINCT user_cn
                           FROM reset_password_tokens
                                    LEFT JOIN user_guid_map ON reset_password_tokens.user_cn = user_guid_map.cn
                           WHERE user_guid_map.cn IS NULL
                           UNION ALL
                           SELECT DISTINCT user_cn
                           FROM set_password_tokens
                                    LEFT JOIN user_guid_map ON set_password_tokens.user_cn = user_guid_map.cn
                           WHERE user_guid_map.cn IS NULL
                       )

If this returns any users before the migration has run then the migration will fail.

After the second release the UpdateUserGUIDMap task should be run on production so that it is up to date with the latest versions.

I will put these directions in the ticket specific release actions upon merge - let me know if anything needs clarifying :) 

I have not updated any further documentation - let me know if there is any that could be relevant!

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/tasks/UpdateUserGUIDMap.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/tasks/UpdateUserGUIDMap.kt
@@ -14,6 +14,7 @@ import kotlin.time.Duration.Companion.minutes
 
 /*
  * Update the user_guid_map table to contain latest up to date userGUIDs and userCNs from Active Directory
+ * Should not be run on production as we will lose record of users deleted from AD
  */
 class UpdateUserGUIDMap(
     private val ldapConfig: LDAPConfig,
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/tasks/UpdateUserGUIDMap.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/tasks/UpdateUserGUIDMap.kt
@@ -3,15 +3,17 @@ package uk.gov.communities.delta.auth.tasks
 import org.jetbrains.annotations.Blocking
 import org.slf4j.LoggerFactory
 import uk.gov.communities.delta.auth.config.LDAPConfig
-import uk.gov.communities.delta.auth.repositories.*
+import uk.gov.communities.delta.auth.repositories.DbPool
+import uk.gov.communities.delta.auth.repositories.LdapRepository
+import uk.gov.communities.delta.auth.repositories.getNewModeObjectGuidString
+import uk.gov.communities.delta.auth.repositories.searchPaged
 import java.sql.Connection
+import java.util.*
 import javax.naming.directory.SearchControls
 import kotlin.time.Duration.Companion.minutes
 
 /*
- * Update the user_guid_map database table by reading all users from Active Directory in both old and new GUID mode.
- * The migration to the new GUID format in Delta is now complete so this job is no longer run regularly,
- * though we'll keep the code and table around for a short while.
+ * Update the user_guid_map table to contain latest up to date userGUIDs and userCNs from Active Directory
  */
 class UpdateUserGUIDMap(
     private val ldapConfig: LDAPConfig,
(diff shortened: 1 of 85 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/tasks/UpdateUserGUIDMap.kt
```diff
@@ -3,60 +3,34 @@ package uk.gov.communities.delta.auth.tasks
...
 import java.sql.Connection
+import java.util.*
 import javax.naming.directory.SearchControls
 import kotlin.time.Duration.Companion.minutes
 
 /*
- * Update the user_guid_map database table by reading all users from Active Directory in both old and new GUID mode.
- * The migration to the new GUID format in Delta is now complete so this job is no longer run regularly,
- * though we'll keep the code and table around for a short while.
+ * Update the user_guid_map table to contain latest up to date userGUIDs and userCNs from Active Directory
```
BenRamchandani, line 16: put comment to never run this on prod again, would lose any users deleted from ad

BenRamchandani (approved):
Should double check the reset password endpoints again for leaking whether a user exists

PR #147, opened 2024-06-11
DT-836 allow CORS for API

JB-softwire:
**Description** _Brief description of the changes. Include screenshots of frontend changes. The PR title should include a ticket number and a few word title/summary._

**Local testing** _If there's anything you haven't tested, aren't confident in, or think is risky note it here. Make sure there are testing instructions on the ticket before merging._

**Release** _Does this change require any [extra release steps](https://dluhcdigital.atlassian.net/wiki/spaces/DT/pages/3375147/Ticket+specific+release+actions), changes to be deployed in a certain order, or affect other devs? Does any documentation need updating?_

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
@@ -5,6 +5,7 @@ import io.ktor.server.application.*
 import io.ktor.server.engine.*
 import io.ktor.server.netty.*
 import io.ktor.server.plugins.cors.routing.*
+import uk.gov.communities.delta.auth.config.Env
 import uk.gov.communities.delta.auth.plugins.configureMonitoring
 import uk.gov.communities.delta.auth.plugins.configureSerialization
 import uk.gov.communities.delta.auth.plugins.configureStatusPages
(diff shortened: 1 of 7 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
```diff
@@ -23,6 +25,11 @@ fun main() {
             port = 8443
         }
         module {
+            install(CORS) {
+                allowHost("api.delta.test.communities.gov.uk")
```
hugh-emerson, line 29: You forgot to use the env var here.

Something like 
```
                allowHost(Env.getOptionalOrDevFallback("API_ORIGIN", "http://localhost:8080")
```

hugh-emerson: It's okay calling it API_ORIGIN because I guess that's what it is, but you could maybe put a code comment here just to say this is to allow access from Swagger.

hugh-emerson (commented):
Regarding the CORS config itself:

* You might also need this in the install block - `method(HttpMethod.Options)`
* Might be nice to specifically configure CORS for just the access token route since that's the only place we need it. But then I think you need to explicitly create a route for the OPTIONS http method. https://ktor.io/docs/server-plugins.html#install-route It's not a huge deal.

hugh-emerson (approved):
👍 Happy for you to merge this and test it via the test env's swagger.

PR #150, opened 2024-06-12
DT-753: Replace GitHub Actions access keys with OIDC

YD-DLUHC:
This PR creates the roles and policies required to use OIDC instead of secrets/keys within GitHub Actions.

### DIFF
--- a/terraform/production/ci_iam.tf
+++ b/terraform/production/ci_iam.tf
@@ -2,7 +2,7 @@ data "aws_caller_identity" "current" {}
 data "aws_region" "current" {}
 
 resource "aws_iam_role" "github_actions_delta_auth_ci" {
-  name = "github-actions-delta-auth-ci"
+  name               = "github-actions-delta-auth-ci"
   assume_role_policy = data.aws_iam_policy_document.github_actions_delta_auth_assume_role.json
 }
 
@@ -26,8 +26,8 @@ data "aws_iam_policy_document" "github_actions_delta_auth_assume_role" {
     }
 
     condition {
-      test     = "StringLike"
-      values   = [
+      test = "StringLike"
+      values = [
         "repo:communitiesuk/delta-auth-service:*"
       ]
       variable = "token.actions.githubusercontent.com:sub"
--- a/terraform/production/ci_iam.tf
+++ b/terraform/production/ci_iam.tf
@@ -30,7 +30,7 @@ data "aws_iam_policy_document" "github_actions_delta_auth_assume_role" {
     }
 
     condition {
-      test = "StringLike"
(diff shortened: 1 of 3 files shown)


hugh-emerson (approved):
LGTM

terraform/production/ci_iam.tf
```diff
@@ -0,0 +1,78 @@
...
+      "ecr:CompleteLayerUpload",
+      "ecr:UploadLayerPart",
+      "ecr:DescribeImages",
+      "ecr:InitiateLayerUpload",
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:PutImage"
+    ]
+
+    resources = [
+      "468442790030.dkr.ecr.eu-west-1.amazonaws.com/delta-auth-service"
```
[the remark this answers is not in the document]
YD-DLUHC: Have changed it to use data.aws_caller_identity.current.account_id if that's what you meant?

terraform/production/ci_iam.tf
```diff
@@ -0,0 +1,78 @@
...
+    condition {
+      test     = "StringEquals"
+      values   = ["sts.amazonaws.com"]
+      variable = "token.actions.githubusercontent.com:aud"
+    }
+
+    condition {
+      test = "StringLike"
+      values = [
+        "repo:communitiesuk/delta-auth-service:*"
```
BenRamchandani, line 31: This should be restricted to the "publish" environment

terraform/production/ci_iam.tf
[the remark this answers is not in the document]
BenRamchandani: That's probably fine. You could make a local var like Hugh suggested

```hcl
locals {
  ecr_repo = "${data.aws_caller_identity.current.account_id}.dkr.ecr.eu-west-1.amazonaws.com/delta-auth-service"
}
```
and then reference it here and where it's used in main.tf

BenRamchandani (changes requested):

terraform/production/ci_iam.tf
YD-DLUHC: changed to restrict to publish

terraform/production/ci_iam.tf
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    condition {
+      test     = "StringEquals"
+      values   = ["sts.amazonaws.com"]
+      variable = "token.actions.githubusercontent.com:aud"
+    }
+
+    condition {
+      test = "StringLike"
```
BenRamchandani, line 33:
```suggestion
      test = "StringEquals"
```

BenRamchandani (approved):

PR #152, opened 2024-06-12
DT-836 remove nonexistent CORS imports

JB-softwire:
Removing stray imports

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
@@ -1,8 +1,11 @@
 package uk.gov.communities.delta.auth
 
+import io.ktor.http.*
 import io.ktor.server.application.*
 import io.ktor.server.engine.*
 import io.ktor.server.netty.*
+import io.ktor.server.plugins.cors.routing.*
+import uk.gov.communities.delta.auth.config.Env
 import uk.gov.communities.delta.auth.plugins.configureMonitoring
 import uk.gov.communities.delta.auth.plugins.configureSerialization
 import uk.gov.communities.delta.auth.plugins.configureStatusPages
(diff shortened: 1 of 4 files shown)


BenRamchandani (approved):
Comments on the CORS change:

* We should send Access-Control-Allow-Methods and -Headers in response to the POST request too. Might not be strictly required, but it's weird they're not the same as OPTIONS
* I don't think you need to explicitly allow the "Content-Type" header
* nit but the description of the api_origin tf variable could be clearer, e.g. "Scheme and domain name for the Delta API Swagger page, used by the auth service to set CORS headers on the API authentication endpoint"
* Not critical, but there's a pattern for having configuration like the API_ORIGIN environment variable in one of the Config files and have it logged at startup, rather than reading it inline

hugh-emerson (changes requested):
See my message on slack. The previous version was very close to working. It'll be better to get it working with the plugin.

You need `allowHeader("X-Requested-With")` but not any other `allowHeader` calls.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Application.kt
```diff
@@ -23,6 +26,11 @@ fun main() {
             port = 8443
         }
         module {
+            install(CORS) {
+                // this is to allow access from Swagger
+                allowHost(Env.getRequiredOrDevFallback("API_ORIGIN", "localhost:8080"))
+                allowHeader("X-Requested-With")
+            }
```
BenRamchandani, line 33: Can you install it to the route rather than globally?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/Routing.kt
```diff
@@ -252,11 +252,13 @@ fun Route.deltaApiRoutes(
     internalDeltaApiTokenController: InternalDeltaApiTokenController,
 ) {
     route("/delta-api/oauth/token") {
+        install(CORS) {
+            // this is to allow access from Swagger
+            allowHost(Env.getRequiredOrDevFallback("API_ORIGIN", "localhost:8080"))
+            allowHeader("X-Requested-With")
```
hugh-emerson, line 258: Did this work? You don't need `allowMethod(HttpMethod.Options`)?

JB-softwire: Running locally both the proper POST request and the OPTIONS request work.

hugh-emerson (approved):
If you've tested it and it works, then 👍

PR #153, opened 2024-06-13
DT-732 AWS X-Ray tracing via OpenTelemetry

BenRamchandani:
**Description**

There are two options for using X-Ray: the X-Ray SDK and OpenTelemetry. The SDK seems more mature, but only supports the servlet API. AWS seems to be encouraging use of OpenTelemetry, although the core library should be stable I've used several integrations which are in alpha.

Add support for OpenTelemetry to the auth service, including running a collector as a sidecar in ECS, all optionally enabled by a Terraform variable and associated feature flag, currently on only for the test environment.

The unit tests log spans to stderr, but otherwise the tracing is disabled locally by default, but can be abled using the feature flag and running the collector with a docker compose profile.

**Local testing** Tested a bit locally and on test. It definitely sends some stuff to AWS X-Ray, though I haven't tested much past that.

**Release** Normal Terraform apply. Feature flagged but should still be sanity tested on staging with the flag off.


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/security/ADLdapLoginService.kt
```diff
@@ -73,6 +75,7 @@ class ADLdapLoginService(
     @Blocking
     private fun ldapBind(userDn: String, password: String): IADLdapLoginService.LdapLoginResult {
         var context: InitialDirContext? = null
+        val span = ldapSpanFactory("AD-ldap-user-bind").startSpan()
         return try {
             context = ldapRepository.bind(userDn, password)
```
hugh-emerson, line 80: Does it make sense to do it inside ldapRepository.bind instead? Or are you deliberately not logging every lookup?

hugh-emerson (approved):

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/security/ADLdapLoginService.kt
BenRamchandani: We'd need another span for `mapUserFromContext` then (my understanding is that's also an LDAP call), and for running the block in `useServiceUserBind`. I think either would be fine but probably not worth changing

BenRamchandani:
One thing I meant to mention: OpenTelemetry can also be used for metrics where we currently use MicroMeter, I don't think there's any rush to replace MicroMeter, especially as the OTel libraries aren't all stable, but something to keep in mind

PR #162, opened 2024-06-27
Add tracing to Azure AD token request

BenRamchandani:
Tested locally:

image


keithkennedy (approved):
I presume this is traced when SSO is used for Microsoft accounts?

YD-DLUHC (approved):

BenRamchandani:
> I presume this is traced when SSO is used for Microsoft accounts?

Yep exactly, that HttpClient is used by the OAuth plugin to do the token request.

PR #165, opened 2024-07-31
DT-1084: User bulk upload access groups ignores comment field

YD-DLUHC:
**Description** Fix for a bug on the bulk update user access groups where any comments added to the file are ignored on the update. This fix modifies the Auth Service endpoint to take in an additional optional comment parameter, which it then uses to check against existing comments and amend as necessary. 

The Readme file has also been updated to add some instructions for when setting up the Auth Service to run locally. 

**Local testing** Have tested multiple times locally using a CSV file and comments are updating correctly. The relevant endpoint within the Auth Service is also used by other services such as the payments approver/reviewer service. I have tested this also and this also works correctly.

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/DeltaUserDetailsRequest.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/DeltaUserDetailsRequest.kt
@@ -120,4 +120,6 @@ data class DeltaUserDetailsRequest(
     @SerialName("organisations") val organisations: List<String>,
     @SerialName("comment") val comment: String? = null,
     @SerialName("classificationType") val classificationType: String? = null, //Not used anywhere yet
+    @SerialName("userObjectGuid") val userObjectGuid: String? = null,
+    @SerialName("isAdmin") val isAdmin: String? = null,
 )
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
@@ -13,10 +13,13 @@ import uk.gov.communities.delta.auth.plugins.ApiError
 import uk.gov.communities.delta.auth.repositories.LdapUser
 import uk.gov.communities.delta.auth.repositories.isInternal
 import uk.gov.communities.delta.auth.services.*
+import uk.gov.communities.delta.auth.utils.getModificationItem
+import javax.naming.directory.ModificationItem
 
 class EditAccessGroupsController(
     private val userLookupService: UserLookupService,
     private val userGUIDMapService: UserGUIDMapService,
+    private val userService: UserService,
     private val groupService: GroupService,
     private val organisationService: OrganisationService,
     private val accessGroupsService: AccessGroupsService,
@@ -26,8 +29,9 @@ class EditAccessGroupsController(
 
     private val logger = LoggerFactory.getLogger(javaClass)
 
-    // This endpoint takes a single user cn, single access group cn and a list of organisation codes.
+    // This endpoint takes a single user cn, single access group cn, a list of organisation codes and a comment.
     // It will assign the user to that access group and the given list of organisations for that access group.
+    // If a comment is posted, it also checks the comments for any modifications and updates them accordingly
     suspend fun addUserToAccessGroup(call: ApplicationCall) {
         val session = call.principal<OAuthSession>()!!
         val callingUser = userLookupService.lookupCurrentUser(session)
@@ -40,6 +44,7 @@ class EditAccessGroupsController(
 
         val targetGroupName = addGroupRequest.accessGroupName
         val targetOrganisationCodes = addGroupRequest.organisationCodes
+        val comment = addGroupRequest.comment
 
         logger.info(
             "Adding user {} to access group {} and organisations {}",
@@ -74,6 +79,15 @@ class EditAccessGroupsController(
                         accessGroup.registrationDisplayName,
                     )
                 }
+            }
+        }
+
+        val commentModifications = comment?.let { getCommentModifications(targetUser, it) }
+
+        if (commentModifications != null) {
+            if (commentModifications.isNotEmpty()) commentModifications.let {
+                userService.updateUser(targetUser,
+                    it, session, call)
             }
         }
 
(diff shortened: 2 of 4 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/DeltaUserDetailsRequest.kt
```diff
@@ -120,4 +120,6 @@ data class DeltaUserDetailsRequest(
     @SerialName("organisations") val organisations: List<String>,
     @SerialName("comment") val comment: String? = null,
     @SerialName("classificationType") val classificationType: String? = null, //Not used anywhere yet
+    @SerialName("userObjectGuid") val userObjectGuid: String? = null,
+    @SerialName("isAdmin") val isAdmin: String? = null,
```
hugh-emerson, line 124: Why was this change necessary when your Delta PR doesn't seem to change anything about sending these fields?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
```diff
@@ -386,6 +400,21 @@ class EditAccessGroupsController(
     private fun getGroupOrgADName(targetGroupName: String, targetOrganisationCode: String?) =
         LDAPConfig.DATAMART_DELTA_PREFIX + targetGroupName + if (targetOrganisationCode.isNullOrEmpty()) "" else "-$targetOrganisationCode"
 
+    private fun getCommentModifications(
+        currentUser: LdapUser,
+        newComment: String
+    ): Array<ModificationItem> {
```
hugh-emerson, line 406: Why is this returning an array of ModificationItem instead of just a nullable ModificationItem? It would simplify this function and also lines 87 to 92.

Line 87 even handles null already, even though this function currently can't return null.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
```diff
@@ -26,8 +29,9 @@ class EditAccessGroupsController(
 
     private val logger = LoggerFactory.getLogger(javaClass)
 
-    // This endpoint takes a single user cn, single access group cn and a list of organisation codes.
+    // This endpoint takes a single user cn, single access group cn, a list of organisation codes and a comment.
     // It will assign the user to that access group and the given list of organisations for that access group.
+    // If a comment is posted, it also checks the comments for any modifications and updates them accordingly
```
hugh-emerson, line 34: It doesn't "check the comment for any modifications". It takes the comment field and if present, appends it to the existing comment field.

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/EditAccessGroupsControllerTest.kt
hugh-emerson, line 1: Can you update an existing test (or add a new one) to verify the comment is updated correctly?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
YD-DLUHC: the code is using the existing userService to apply the update which is currently wired to take in an array of ModificationItem. If I change my code to use a single item then from what I can see I'd need to modify/create three more functions (updateUser, updateUserOnAD, getAuditData) to take the new parameter, unless there's an easier way. Shall I leave as is or change/create new functions?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
YD-DLUHC: if (newComment.isNotEmpty() && newComment != currentUser.comment) - I'd have thought this is technically comparing the two?

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/DeltaUserDetailsRequest.kt
YD-DLUHC: Forgot to remove this from when I tried to send the userDetails field from Delta along with the comment

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
hugh-emerson: You could do put it in an array on the line where you call userService, like `userService.updateUser(targetUser, arrayOf(commentModification), session, call)`
Still simplifies this class, e.g. this function doesn't need to look like it can create multiple modification items. 

```
if (newComment.isNotEmpty() && newComment != currentUser.comment) {
       val updatedComment = "${currentUser.comment}\n$newComment"
       return getModificationItem("comment", currentUser.comment, updatedComment)
}
return null
```

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
hugh-emerson: Yeah but I'm not sure `newComment != currentUser.comment` will ever really be relevant since the comment gets appended and doesn't replace the old comment. Even if the comment has only ever been set via bulk upload and you upload the same file twice, it's actually still going to update the comment.

Say the comment is "foobar" in the spreadsheet. First upload would set comment to "\nfoobar" and then second upload would set it to "\nfoobar\nfoobar". But I don't think this is worth worrying about or even a bad thing necessarily. It's fine to keep appending to the comment so they can see that the user was updated without digging through the audit trail.

Practically speaking, we're just always appending the comment and I think that's fine. That's what the ticket asked for.

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/internal/EditAccessGroupsController.kt
YD-DLUHC: Changed to the suggested code

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/EditAccessGroupsControllerTest.kt
YD-DLUHC: Added tests for adding new comments and appending to existing comments

PR #178, opened 2024-10-21
Dt 1146

hugh-emerson:
**Description**
* Add in stats data certifier as a role that can be returned
* Add in/update tests
* In the process I ended up refactoring a lot...


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/repositories/LdapRepository.kt
```diff
@@ -243,26 +238,22 @@ class LdapRepository(
...
+        val allGroups = user.memberOfCNs
+        val userRoles = allGroups.mapNotNull { role ->
             when {
-                role.contains("${datamartDeltaDataProviders}$organisationId", ignoreCase = true) -> "Data provider"
-                role.contains("${datamartDeltaDataCertifiers}$organisationId", ignoreCase = true) -> "Data certifier"
-                else -> role
+                role.equals("${datamartDeltaDataProviders}$organisationId", ignoreCase = true) -> "Data provider"
+                role.equals("${datamartDeltaDataCertifiers}$organisationId", ignoreCase = true) -> "Data certifier"
+                role.equals("${datamartDeltaStatsDataCertifiers}$organisationId", ignoreCase = true) -> "Data certifier"
+                else -> null
```
YD-DLUHC, line 248: Just checking - are both certifier types the same? i.e. both would show as 'Data certifier'?

YD-DLUHC (approved):
LGTM

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/repositories/LdapRepository.kt
hugh-emerson: They're slightly different but I don't think we need to distinguish between them here.

PR #180, opened 2024-10-25
DT-1277

Mohsina-Ahmed:
Updated new coat if arms.

**Description** 
Updated new coat of arms. Added new css files and new images


auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
```diff
@@ -17,7 +17,7 @@
     <link href="/static/assets/images/govuk-apple-touch-icon-152x152.png" rel="apple-touch-icon"
           sizes="152x152">
     <link href="/static/assets/images/govuk-apple-touch-icon.png" rel="apple-touch-icon">
-    <link href="/static/govuk-frontend-4.8.0.min.css" rel="stylesheet">
```
hugh-emerson, line 20: And should govuk-frontend-4.8.0 be deleted from the repo?

hugh-emerson (changes requested):

auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
Mohsina-Ahmed: @hugh-emerson 
Ideally yes. I don't think we need that old file anymore. I will do that now.

hugh-emerson (approved):
👍

PR #184, opened 2024-11-22
allow for longer top level domains

kierangarside09:
**Description** 
Changed limit of TLD from 2-4 to have no limit due to issue with .wales TLDs

**Local testing**
N/A

**Release** 
Kotlin


hugh-emerson (approved):
Try to put the ticket number in the commit message, even if it's MSD rather than DT. But fine to merge this anyway 👍 
Just bump the `VERSION` in publish.yml and we can deploy this on Monday.

PR #186, opened 2024-12-11
DT-1361: Comments not being added when performing bulk update access …

YD-MHCLG:
**Description** 

Fix for a bug on the bulk update user access groups where any comments added to the file are ignored on the update when removing access groups. This fix modifies the Auth Service endpoint to take in an additional optional comment parameter, which it then uses to check against existing comments and amend as necessary.

The changes required on auth-service have been made [in this PR](https://github.com/communitiesuk/delta/pull/1784)

**Local testing** 

Have tested multiple times locally using a CSV file and comments are updating correctly. 

- Find a user to test with.

- Go to Admin>Edit user. 

- Find an access group to remove from the user and make a note of the org code/org name/access group name.

- Open a new tab onto Admin>Users>Bulk Actions>Update user access groups.

- Either create a new file to upload or use the file here. The file will need to contain the information needed to remove an access group and organisation from an existing user. Ensure the comment field is filled in. Click save and await the 'success’ message. 

- Go back to the Admin>Users>User details tab previously opened and refresh the page. The access group should have been removed from the user along with the comment from the CSV file.


OliviaKRose (approved):
lgtm 🚢

PR #189, opened 2025-02-03
updated README with db username and password

OliviaKRose:
**Description**
Updated README with db username and password for local auth service db

### DIFF
--- a/auth-service/README.md
+++ b/auth-service/README.md
@@ -8,6 +8,7 @@
   and delta for port forwarding commands
 * Copy `.env.template` to `.env` and fill in the uncommented variables as instructed in that file
 * Postgres, by default on port 5438, use `docker compose up -d`
+    * When connecting to the local db, both the username and password are `postgres`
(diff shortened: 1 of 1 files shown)


auth-service/README.md
```diff
@@ -8,6 +8,7 @@
   and delta for port forwarding commands
 * Copy `.env.template` to `.env` and fill in the uncommented variables as instructed in that file
 * Postgres, by default on port 5438, use `docker compose up -d`
+    * When connecting to the local db, both the username and password are `postgres`
```
keithkennedy, line 11: May be worth linking to the source code - in case we ever changed it, too.

https://github.com/communitiesuk/delta-auth-service/blob/852362da13c738189cc4783b963297813e54a9f4/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/config/DatabaseConfig.kt

keithkennedy (approved):

auth-service/README.md
OliviaKRose: added pr here to add link https://github.com/communitiesuk/delta-auth-service/pull/190

PR #192, opened 2025-02-07
DT-1374 Add welcome message for new internal users

OliviaKRose:
**Description**
Added check to see if a user is new by check the audit logs to see how many times they have logged in. This is then passed to Delta

**Local testing** 
Tested locally
Added unit tests (also added clean up to some pre-existing tests - only for user-audit and user_guid_map tables)

**Release**
Also requires Delta release

### DIFF
--- a/auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/RefreshUserInfoControllerTest.kt
+++ b/auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/RefreshUserInfoControllerTest.kt
@@ -36,6 +36,8 @@ class RefreshUserInfoControllerTest {
 
     @Test
     fun testUserInfoEndpoint() = testSuspend {
+        coEvery { userAuditService.checkIsNewUser(user.getGUID()) }.returns(false)
+
         testClient.get("/user-info") {
             headers {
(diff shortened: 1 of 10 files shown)


hugh-emerson (commented):
LGTM, but is there a relevant test to update?

You also need to get the pipeline passing. Is that why it's a draft PR?

OliviaKRose:
> LGTM, but is there a relevant test to update?
> 
> You also need to get the pipeline passing. Is that why it's a draft PR?

Yes, it's in draft as I'm still working on the tests plus I need to get the pipeline passing

auth-service/src/test/kotlin/uk/gov/communities/delta/controllers/RefreshUserInfoControllerTest.kt
```diff
@@ -36,6 +36,8 @@ class RefreshUserInfoControllerTest {
 
     @Test
     fun testUserInfoEndpoint() = testSuspend {
+        coEvery { userAuditService.checkIsNewUser(user.getGUID()) }.returns(false)
```
hugh-emerson, line 39: Should you add a corresponding `assertEquals` below?

hugh-emerson (approved):
One small comment, but merge when you're ready.

PR #193, opened 2025-02-07
DT-1479: added additional log for login failure type for api token request

keithkennedy:
**Description** 
Added additional log for login failure type for api token request

### DIFF
--- a/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/ExternalDeltaApiTokenController.kt
+++ b/auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/ExternalDeltaApiTokenController.kt
@@ -56,6 +56,12 @@ class ExternalDeltaApiTokenController(
         val userCn = username.replace('@', '!')
         val loginResult = ldapService.ldapLogin(userCn, password)
         if (loginResult !is IADLdapLoginService.LdapLoginSuccess) {
+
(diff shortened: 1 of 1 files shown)


auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/ExternalDeltaApiTokenController.kt
```diff
@@ -56,6 +56,12 @@ class ExternalDeltaApiTokenController(
         val userCn = username.replace('@', '!')
         val loginResult = ldapService.ldapLogin(userCn, password)
         if (loginResult !is IADLdapLoginService.LdapLoginSuccess) {
+
+            logger.atInfo()
+                .addKeyValue("userCn", userCn)
```
hugh-emerson, line 61: Or use they key "username" to match other logs? Though I guess those ones do the version with `!` instead of `@`?

hugh-emerson (approved):

auth-service/src/main/kotlin/uk/gov/communities/delta/auth/controllers/external/ExternalDeltaApiTokenController.kt
keithkennedy: Agree, should be the same as the previous log for searching purposes.
I added the client ID to the new log, too.

PR #197, opened 2025-03-21
DT-1502: non-prod warning text

keithkennedy:
**Description** 
Added warning that should only appear in non-prod environments
Added noindex to `<head>` that should apply in non-prod environments (feel like we should have this for prod, too?)

**Local testing** 
- Locally tested with switching variable manually
- Unit test coverage

**Release** 
Auth service release

### DIFF
--- a/auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
+++ b/auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
@@ -42,7 +42,7 @@
 
 <div th:fragment="environment-warning-message" class="">
     <div th:unless="${isProduction}">
-        <div class="govuk-width-container" style="margin-top: 10px">
+        <div class="govuk-width-container govuk-!-margin-top-2">
             <div class="govuk-notification-banner" role="region" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
                 <div class="govuk-notification-banner__header">
                     <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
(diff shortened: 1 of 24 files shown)


auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
```diff
@@ -39,6 +40,28 @@
...
+                    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
+                        Important
+                    </h2>
+                </div>
+                <div class="govuk-notification-banner__content">
+                    <p class="govuk-notification-banner__heading">
+                        This is a staging site, used for internal testing.<br>Do not enter data.
+                    </p>
+                    <p class="govuk-body">
+                        To enter data, go to the <a class="govuk-notification-banner__link" href="https://delta.gov.uk">main Delta site</a>.
```
hugh-emerson, line 57: The href is wrong

auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
```diff
@@ -1,12 +1,13 @@
...
 <html class="govuk-template " lang="en" xmlns:th="http://www.thymeleaf.org">
-
+<!--/*@thymesVar id="isProduction" type="java.lang.Boolean" */-->
 <head th:fragment="head (title)">
     <meta charset="utf-8">
     <title th:text="${title}">Delta</title>
     <meta content="width=device-width, initial-scale=1, viewport-fit=cover" name="viewport">
     <meta content="#0b0c0c" name="theme-color">
     <meta content="IE=edge" http-equiv="X-UA-Compatible">
+    <meta th:unless="${isProduction}" name="robots" content="noindex">
```
hugh-emerson, line 10: I think you're correct to only do this for non-production because evidently people use google to find the site, which is fine.

But you also need to add this to the Delta website because there are public pages.

auth-service/src/main/resources/templates/thymeleaf/fragments/template.html
keithkennedy: Thanks, fixed now.

hugh-emerson (approved):

PR #199, opened 2025-03-28
DT-1558: Broken hyperlink in the No User Account email

YD-MHCLG:
**Description** 

[DT-1558](https://mhclgdigital.atlassian.net/browse/DT-1558) - Fixes a broken URL in an email template

**Local testing** 

Tested locally and the new link works fine

[DT-1558]: https://mhclgdigital.atlassian.net/browse/DT-1558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/auth-service/src/main/resources/templates/thymeleaf/emails/no-user-account.html
+++ b/auth-service/src/main/resources/templates/thymeleaf/emails/no-user-account.html
@@ -17,7 +17,7 @@
 <p>If this was you please check this is the email address linked to your account or click the link below to create an
     account.</p>
 <p>
-    <a th:href="@{${deltaUrl}+'/delta/register'}">Create an account</a>.
+    <a th:href="@{${authServiceUrl}+'/delta/register'}">Create an account</a>.
 </p>
 <p>If this wasn't you, no action is required.</p>
 <p>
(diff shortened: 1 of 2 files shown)


auth-service/src/main/resources/templates/thymeleaf/emails/no-user-account.html
```diff
@@ -17,11 +17,11 @@
 <p>If this was you please check this is the email address linked to your account or click the link below to create an
     account.</p>
 <p>
-    <a href="/delta/register">Create an account</a>
+    <a th:href="@{${deltaUrl}+'/delta/register'}">Create an account</a>.
```
keithkennedy, line 20: Is the `deltaUrl` equal to (in test), `https://delta.test.communities.gov.uk` ?
If so, this would end up with the URL: `https://delta.test.communities.gov.uk/delta/register`, which is invalid.
We should end up with auth service URL + `/delta/register/` -> `https://auth.delta.test.communities.gov.uk/delta/register`

keithkennedy (changes requested):

keithkennedy (approved):

